PR #111, opened 2014-11-12
discover scheduler should wait full timeout for severs to respond

HenryMiller1:
Fix for issue 110.  It is believed that this also fixes issue 107, but I can't test that here.  

Note that I changed the timeout parameter from milliseconds to seconds.


HenryMiller1:
I messed up my local repository by not properly using branches.  Closing this to get a clean fix.

Issue #79, opened 2013-12-27
OSX 10.9 mavericks and xcode 5: icecc --build-native uses wrong paths for some tools, fails


ec1oud:
$ icecc --build-native
adding file /bin/true=/usr/bin/true
adding file /usr/lib/libSystem.B.dylib
adding file /usr/lib/system/libcache.dylib
adding file /usr/lib/system/libsystem_pthread.dylib
adding file /usr/lib/system/libsystem_kernel.dylib
adding file /usr/lib/system/libsystem_platform.dylib
adding file /usr/lib/system/libdyld.dylib
adding file /usr/lib/system/libsystem_malloc.dylib
adding file /usr/lib/system/libcompiler_rt.dylib
adding file /usr/lib/system/libunwind.dylib
adding file /usr/lib/system/libsystem_c.dylib
adding file /usr/lib/system/libsystem_m.dylib
adding file /usr/lib/system/libdispatch.dylib
adding file /usr/lib/system/libsystem_blocks.dylib
adding file /usr/lib/system/libsystem_asl.dylib
adding file /usr/lib/system/liblaunch.dylib
adding file /usr/lib/system/libquarantine.dylib
adding file /usr/lib/system/libmacho.dylib
adding file /usr/lib/system/libcorecrypto.dylib
adding file /usr/lib/system/libxpc.dylib
adding file /usr/lib/system/libsystem_sandbox.dylib
adding file /usr/lib/libobjc.A.dylib
adding file /usr/lib/libauto.dylib
adding file /usr/lib/libc++.1.dylib
adding file /usr/lib/libc++abi.dylib
adding file /usr/lib/libDiagnosticMessagesClient.dylib
adding file /usr/lib/system/libsystem_notify.dylib
adding file /usr/lib/system/libsystem_info.dylib
adding file /usr/lib/system/libsystem_dnssd.dylib
adding file /usr/lib/system/libsystem_network.dylib
adding file /usr/lib/system/libkeymgr.dylib
adding file /usr/lib/system/libcommonCrypto.dylib
adding file /usr/lib/system/libcopyfile.dylib
adding file /usr/lib/system/libremovefile.dylib
adding file /usr/lib/system/libsystem_configuration.dylib
adding file /usr/lib/system/libsystem_stats.dylib
adding file /usr/lib/system/libunc.dylib
adding file /usr/bin/gcc
adding file /usr/lib/libxcselect.dylib
[...]

PR #118, opened 2015-02-12
Fix the creation of a native environment on MacOS (#79)

clopez:
* Clang is the default compiler on MacOS, so "icecc --build-native"
    should default to build a Clang toolchain on MacOS.

  * Fix on icecc-create-env the detection of the path for gcc.

### DIFF
--- a/client/main.cpp
+++ b/client/main.cpp
@@ -60,6 +60,7 @@
 #include <sys/wait.h>
 
 #include "client.h"
+#include "platform.h"
 
 using namespace std;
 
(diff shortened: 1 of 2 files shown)


client/main.cpp
```diff
@@ -60,6 +60,7 @@
 #include <sys/wait.h>
 
 #include "client.h"
+#include "platform.h"
```
coolo: I don't have neither platform.h nor determine_platform on linux

clopez: On 12/02/15 14:28, Stephan Kulow wrote:
>> @@ -60,6 +60,7 @@
>>  #include <sys/wait.h>
>>  
>>  #include "client.h"
>> +#include "platform.h"
> 
> I don't have neither platform.h nor determine_platform on linux
> 
> ---
> Reply to this email directly or view it on GitHub:
> https://github.com/icecc/icecream/pull/118/files#r24579917
> 

platform.h is services/platform.h

I tested it on Linux, works fine.

??

coolo:
that happens if you visit code only  once a year :)

PR #120, opened 2015-02-12
services: use PATH_MAX instead of FILENAME_MAX

pinotree:
In a Linux-specific code, use `PATH_MAX` instead of `FILENAME_MAX` as buffer size for `getcwd()`.

### DIFF
--- a/services/logging.cpp
+++ b/services/logging.cpp
@@ -25,6 +25,7 @@
 #include "logging.h"
 #include <fstream>
 #include <signal.h>
+#include <limits.h>
 #ifdef __linux__
 #include <dlfcn.h>
 #endif
@@ -63,7 +64,7 @@ void setup_debug(int level, const string &filename, const string &prefix)
 #ifdef __linux__
 
         if (fname[0] != '/') {
-            char buf[FILENAME_MAX];
+            char buf[PATH_MAX];
 
             if (getcwd(buf, sizeof(buf))) {
                 fname.insert(0, "/");


services/logging.cpp
```diff
@@ -63,7 +64,7 @@ void setup_debug(int level, const string &filename, const string &prefix)
 #ifdef __linux__
 
         if (fname[0] != '/') {
-            char buf[FILENAME_MAX];
+            char buf[PATH_MAX];
```
HenryMiller1: After looking up PATH_MAX and FILENAME_MAX: both are incorrect.  linux imposes no limit on path lenghts and thus an exact constant cannot be defined.  The only correct thing a buffer (with malloc) and keep looping on getcwd with larger buffers until it succeeds.

FILENAME_MAX is too large for the stack in general though, so we shouldn't use it. (since we are single threaded I don't think this objection applies to us)

HenryMiller1:
Seems like this should be rolled into #119 which is a better solution to the problem.

ossilator:
PATH_MAX is so large on linux (4k, coming from linux/limits.h) that it isn't an actual concern.

HenryMiller1:
We support linux, *BSD, and OSX. This is mostly a theoretical issue on linux, but for arbitrary platforms it might not be.

HenryMiller1:
That said, the code is in a #if __linux__ block and PATH_MAX seems large enough.  (client/util.h is not building into everything that logging/* is in so get_cwd() needs to move which doesn't seem worth it)

PR #121, opened 2015-02-14
Few cleanups of unneeded configure headers/functions checks

pinotree:
Remove configure checks for headers and functions which are
- not used at all
- POSIX (and thus they ought to be available)
- not POSIX, but are so far used unconditionally

This series should cause no change in the supported OSes, since all the headers and functions previously checked and used continue to be handled as before.

### DIFF
--- a/client/local.cpp
+++ b/client/local.cpp
@@ -32,9 +32,7 @@
 #include <string.h>
 #include <errno.h>
 #include <vector>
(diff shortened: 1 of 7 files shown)


pinotree:
@HenryMiller1 could you please take a look at this? @rbelem basically disappeared...

PR #123, opened 2015-03-23
Avoid accidental assignments

doomdavve:
Fixes recent regression with '~/src/icecream/client/icecc
--build-native gcc' which broke icecc on non-Darwin entirely.

### DIFF
--- a/client/main.cpp
+++ b/client/main.cpp
@@ -175,8 +175,8 @@ static int create_native(char **args)
     if (machine_name.find("Darwin") == 0)
         is_clang = true;
     // Args[0] may be a compiler or the first extra file.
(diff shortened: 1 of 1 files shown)


coolo:
the assignment is not accidental. If it breaks another use case, this another bug. But the assignments are real.

doomdavve:
Sorry about the terse comment, I had a bus to catch...  Ok, I see. Before https://github.com/icecc/icecream/pull/118 it was more clear what was going on.

Well, while that style of assignment works when setting to true, it doesn't work when setting to false, since that will evaluate to false, meaning for gcc, the extrafiles counter will not be incremented for gcc, and 'gcc' will be passed along as an extra file, wrecking havoc later on.

I'll drop this and make a better fix tomorrow.

PR #124, opened 2015-03-24
Fix creation of native environment for gcc

doomdavve:
In 96552a293a55d15d7179285a70868f0a55c9a6ea, the default compiler for
MacOS was changed and because of that the code could no longer assume
that gcc is the true unless clang is explicitly specified. For that
reason the commit explicitly set 'is_clang' to false when 'gcc' is
requested as compiler.

However it did so inside an if expression causing the expression to
evaluate to false always and without incrementing the extrafiles
pointer. This in turn caused the rest of create_native to believe gcc
is an extra file to be added to the tarball. No such file normally
exists so the creation of a native environment failed.

Fix this issue by moving the is_clang assignments outside the if
expressions, making the code slightly longer but clearer.

### DIFF
--- a/client/main.cpp
+++ b/client/main.cpp
@@ -175,8 +175,11 @@ static int create_native(char **args)
     if (machine_name.find("Darwin") == 0)
         is_clang = true;
     // Args[0] may be a compiler or the first extra file.
-    if (args[0] && ((!strcmp(args[0], "clang") && (is_clang = true))
-                    || (!strcmp(args[0], "gcc") && (is_clang = false)))) {
+    if (args[0] && !strcmp(args[0], "clang")) {
+        is_clang = true;
+        extrafiles++;
+    } else if (args[0] && !strcmp(args[0], "gcc")) {
+        is_clang = false;
         extrafiles++;
     }
 


doomdavve:
New version of the misguided https://github.com/icecc/icecream/pull/123. Hopefully this makes it clearer what is going on.

doomdavve:
Ping

coolo:
I don't know about rbelem, but we're still looking for an active maintainer.

doomdavve:
Rodrigo, do you think you can merge this? We use icecc to build for android so we need to use something newer than the 1.0-branch branch. master is quite useless for us now because of the issue this PR fixes.

doomdavve:
After looking into this again, this is how to reproduce the error on my machine:

drogon:~ % ~/pkg/icecc/bin/icecc --build-native gcc
adding file /bin/true
adding file /lib/x86_64-linux-gnu/libc.so.6
adding file /lib64/ld-linux-x86-64.so.2
adding file /usr/bin/gcc
adding file /lib/x86_64-linux-gnu/libm.so.6
adding file /usr/bin/g++
adding file /usr/bin/cc1=/usr/lib/gcc/x86_64-linux-gnu/4.9/cc1
adding file /usr/lib/x86_64-linux-gnu/libcloog-isl.so.4
adding file /usr/lib/x86_64-linux-gnu/libisl.so.15
adding file /usr/lib/x86_64-linux-gnu/libmpc.so.3
adding file /usr/lib/x86_64-linux-gnu/libmpfr.so.4
adding file /usr/lib/x86_64-linux-gnu/libgmp.so.10
adding file /lib/x86_64-linux-gnu/libdl.so.2
adding file /lib/x86_64-linux-gnu/libz.so.1
adding file /usr/bin/cc1plus=/usr/lib/gcc/x86_64-linux-gnu/4.9/cc1plus
adding file /usr/bin/as
adding file /usr/lib/x86_64-linux-gnu/libopcodes-2.25.90-system.20151209.so
adding file /usr/lib/x86_64-linux-gnu/libbfd-2.25.90-system.20151209.so
adding file /usr/lib/gcc/x86_64-linux-gnu/4.9/liblto_plugin.so
ls: cannot access gcc: No such file or directory
adding file gcc=
adding file /etc/ld.so.conf=/tmp/icecc_ld_so_conf2hWyQw
cp: cannot stat ‘gcc=’: No such file or directory
/usr/bin/md5sum: /tmp/iceccenvosBiqR/cc=: No such file or directory
creating 8eabecdce9c77d003a4312be62acb6b0.tar.gz
tar: cc=: Cannot stat: No such file or directory
tar: Exiting with failure status due to previous errors
Couldn't create archive

As long as you don't explicitly have to select gcc, you should be fine. But on mac, where clang is default, the gcc path is broken.

coolo:
closing as per comment in 144

PR #129, opened 2015-04-20
[not-ready] scheduler should not assume environment is on local server

HenryMiller1:
The problem is simple the scheduler assumes that the host the job is coming from has all the environments of the job, but in fact this is not true, in the case of a chroot the host may not have access to the environment (at least not by the paths given)

### DIFF
--- a/scheduler/scheduler.cpp
+++ b/scheduler/scheduler.cpp
@@ -464,10 +464,6 @@ static bool handle_local_job_done(CompileServer *cs, Msg *_m)
    platform).  */
 static string envs_match(CompileServer *cs, const Job *job)
 {
-    if (job->submitter() == cs) {
-        return cs->hostPlatform();    // it will compile itself
-    }
-
     Environments compilerVersions = cs->compilerVersions();
 
     /* Check all installed envs on the candidate CS ...  */


HenryMiller1:
This should fix issue 128

coolo:
the *client* compiles the job if localhost is assigned

HenryMiller1:
Maybe it should, but it is not - my builds were failing.  (there might be a different issue?  if so I'm not sure what it might be)

HenryMiller1:
That is the daemon on localhost is attempting to compile, not the client that issued the job. The daemon does not have the ability to compile this job.

If the client should compile, then there is a bug elsewhere - I'm not sure how that should work.

coolo:
that's clearly a bug somewhere else. the scheduler is correct to ignore the host's env - the client asks the scheduler (indirectly) for the host to use, the scheduler picks 'localhost' and the client will then compile it locally within the chroot. There is no need to transfer an environment from within chroot to the outside. So if that fails, then either the client didn't figure itself it's picked - or there is some environment check at the wrong place.

HenryMiller1:
I'm not sure if I'm not understanding where you think the bug is, or if I'm not explaining why the environment is not accessible on local host.

We are building in a chroot. (actually docker), mounting /var/run/icecc within the chroot to ensure that within the chroot we can reach the daemon running outside the chroot. We then use ICECC_VERSION to point to the correct toolchain environment.  When the scheduler says build on localhost the daemon wants to start a new chroot.  The daemon on localhost looks for ICECC_VERSION - but that file path is relative to the chroot and not the local file system, and thus the local daemon cannot find the file.  The local daemon needs to get the environment from the client across the socket.

The obvious fix is what I put here - the scheduler already tells the client when to transfer to the environment to the daemon.

At least that is my analysis, if I'm wrong can you suggest where the fix should be?

coolo:
it might have changed in the mean time, but when the code you want to remove was written the daemon would *not* compile if the job was coming from localhost. It would have blocked one slot, but let the client in the chroot do the actual compilation. Without preprocessing/transfering overhead - just plain local compilation.

PR #132, opened 2015-05-06
Better error messages on client

HenryMiller1:
I'm getting tired of people asking me "what does exception 17 mean".  Even though icecc retries, seeing a long string of these errors - and associated build slow down - makes users concerned.  My only recourse is to figure out which version they are using, and then dig through the source code for that version to figure it out.

Generally C++ best practice is to only throw something derived from std::exception.  This change switching from throwing int to throwing std::exception, and then uses exception.what() to give the user a better error message.

### DIFF
--- a/client/client.h
+++ b/client/client.h
@@ -30,6 +30,8 @@
 #include <sys/time.h>
 #include <sys/resource.h>
 
+#include <stdexcept>
+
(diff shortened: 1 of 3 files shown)


coolo:
yeah, that's an heritage from distcc that was good enough for a long time :)

PR #135, opened 2015-07-23
Retrieve the physical number of cores not counting hyperthreaded cores

permcody:
Modern MacPro workstations have 12 physical cores which are hyperthreaded. The current npcus logic returns 24 in this case which is a bit high if you are going for decent load balancing. This change will only report physical CPUs which makes the whole network run better.

### DIFF
--- a/daemon/ncpus.c
+++ b/daemon/ncpus.c
@@ -115,10 +115,8 @@ int dcc_ncpus(int *ncpus)
 {
     int mib[2];
     size_t len = sizeof(*ncpus);
-    mib[0] = CTL_HW;
-    mib[1] = HW_NCPU;
 
-    if (sysctl(mib, 2, ncpus, &len, NULL, 0) == 0) {
+    if (sysctlbyname("hw.physicalcpu", ncpus, &len, NULL, 0) == 0) {
         return 0;
     }
 


ossilator:
i find this rather questionable. can you cite any references which recommend this approach?

coolo:
I would think it actually runs faster with 24 than with 12 if you have the memory to go with it - they are hyperthreading for a reason. Of course only if you max out the full farm.

But I would also be worried that the call you're doing is portable enough to cover all the OSes.

permcody:
I don't have any references but we have around 8 Mac Pros in our farm. If we do make -j80 or -j100 all the jobs end up on just 3-4 systems, at least initially until the second round of files start trickling out for compilation. With this patch the load gets spread out more evenly over all the systems. I agree that hyper-threading does indeed help but usually not all the way to 2x (24 vs 12).

I'll see if I can get some timing numbers and report those back. It'll be a few days. If you are against this, how about a patch that allows the users to select whether to count cores or hyprethreads when the daemon launches?

BTW - This is a cool piece of software! It's a little rough on the Macs so my team and I may submit other patches to fix things up. Thanks

permcody:
As to the portability question, I'll do a little digging and see what I can find. This call only happens on the BSD and Mac systems so it won't break the other Linux distros. We could always break this line out into it's own section (Macs only) if it turns out that that API is not available elsewhere.

ossilator:
it's true that distributing over physical cores first and over the secondary threads only when all first threads are busy would yield more performance, irrespective of memory size and bandwidth of the machines. i would consider such a patch, but it would be a lot more complex than what you created now, especially when different machines are involved (consider that the scheduler estimates the speed of machines). additionally, the operating system may thwart your attempts at using the cores evenly depending on the power saving strategy.

just cutting down the number of used threads per machine will reduce the total capacity of the farm, which will affect performance if multiple jobs run concurrently, which is why the patch is not acceptable as is. still, if you want to do it that way, you can just manually limit the number of threads each node accepts.

friedmud:
@ossilator how do you manually limit the number of threads a node accepts?

friedmud:
Nevermind - found it.

permcody:
We will limit the number of threads a box can accept for now. I think there are bigger improvements to the scheduler that could be made in the future to improve performance.

Closing this PR

Issue #86, opened 2014-03-19
g++: error trying to exec 'objcopy': execvp: No such file or directory


ds-hwang:
Following option of gcc 4.8 uses objcopy, but icecc-create-env does not handle objcopy
`'cflags': ['-gsplit-dwarf'], 'ldflags': ['-Wl,--gdb-index'],`
- See more detail for the build option http://gcc.gnu.org/wiki/DebugFission

So compile fails with following build error

```
[60/10095] CXX obj/third_party/angle/src/compiler/preprocessor/preprocessor.MacroExpander.o
[...]

PR #144, opened 2016-03-14
implement -gsplit-dwarf (dwarf fission) support

johnmiked15:
A discussion about the possibility of `-gsplit-dwarf` support feature request is in #86. For more details about the functionality, please see https://gcc.gnu.org/wiki/DebugFission.

This pull request adds full distributed compilation support when using the `-gsplit-dwarf` flag for both gcc and clang.

There is an oddity with how the compilers embed the `.dwo` file path information into the object file, and is based on the specified path to the output file on the invoked command line. Some utility methods (shamelessly adapted from various Stack Overflow discussions) aids path manipulation without `boost::filesystem`. These path manipulation techniques allow the daemon to recreate the directory structure present on the client, thus tricking the compiler into embedding `.dwo` path information that will match what the client's machine expects.

Tests have been added, though I had to resort to comparing `readelf` output whilst ignoring some debugging specific information since certain tags (`DW_AT_GNU_dwo_id`, for instance) are different on each run. `DW_AT_GNU_dwo_name` and `DW_AT_comp_dir` will be slightly different based on whether it was compiled remotely or locally, as well, since the path manipulation trick is not necessary for local builds.

### DIFF
--- a/client/main.cpp
+++ b/client/main.cpp
@@ -176,7 +176,7 @@ static int create_native(char **args)
         is_clang = true;
     // Args[0] may be a compiler or the first extra file.
     if (args[0] && ((!strcmp(args[0], "clang") && (is_clang = true))
-                    || (!strcmp(args[0], "gcc") && (is_clang = false)))) {
+                    || (!strcmp(args[0], "gcc") && !(is_clang = false)))) {
         extrafiles++;
     }
 
--- a/client/arg.cpp
+++ b/client/arg.cpp
@@ -92,6 +92,7 @@ bool analyse_argv(const char * const *argv, CompileJob &job, bool icerun, list<s
 {
     ArgumentsList args;
     string ofile;
+    string dwofile;
 
 #if CLIENT_DEBUG > 1
     trace() << "scanning arguments ";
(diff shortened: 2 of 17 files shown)


coolo:
That's an impressive work and I can't see anything wrong with it. I can merge it, but I would prefer giving over maintainership of the repo to you. You seem to understand the code well enough to be the maintainer of it :)

johnmiked15:
Oh wow. Let me consider taking maintainership a little bit first but go ahead and merge for now if you feel comfortable with it. Thanks for taking a look!

johnmiked15:
I just pushed on more commit as I forgot to include the size of the dwo file in the job statistics. Let me know if you'd like me to address anything else before merging.

client/main.cpp
```diff
@@ -176,7 +176,7 @@ static int create_native(char **args)
         is_clang = true;
     // Args[0] may be a compiler or the first extra file.
     if (args[0] && ((!strcmp(args[0], "clang") && (is_clang = true))
-                    || (!strcmp(args[0], "gcc") && (is_clang = false)))) {
+                    || (!strcmp(args[0], "gcc") && !(is_clang = false)))) {
```
doomdavve: If this is merged you may close https://github.com/icecc/icecream/pull/124 since it contains essentially this fix. Your way of fixing it is terser and probably readable enough for the audience anyway.

darrengarvey:
From a cursory reading of the changeset, it looks like this could be extended / generalised to add support for precompiled headers. Currently icecream just decides to compile any files that need PCH on the local machine rather than sending across the PCH files similarly to this.

PR #147, opened 2016-03-29
Make --save-temps force local building

kito-cheng:
--save-temps is alias of -save-temps, but some build system will use this one, such as gcc


kito-cheng:
ping

coolo:
thanks for the ping - I was on vacation :)

PR #148, opened 2016-04-11
allow execution of local jobs when daemon's max_jobs is set to 0

hudde:
Starting the daemon with '-m 0' should allow linking to happen locally, but all compile jobs should be executed remotely. Before the compilation process would hang when the scheduler would assign a job for the local daemon with max_jobs set to 0 and the job would never get executed.

### DIFF
--- a/daemon/main.cpp
+++ b/daemon/main.cpp
@@ -1210,7 +1210,8 @@ bool Daemon::handle_job_done(Client *cl, JobDoneMsg *m)
 
 void Daemon::handle_old_request()
 {
-    while ((current_kids + clients.active_processes) < max_kids) {
+    while ((current_kids + clients.active_processes) < max_kids ||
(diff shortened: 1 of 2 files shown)


daemon/main.cpp
```diff
@@ -1210,7 +1210,8 @@ bool Daemon::handle_job_done(Client *cl, JobDoneMsg *m)
 
 void Daemon::handle_old_request()
 {
-    while ((current_kids + clients.active_processes) < max_kids) {
+    while ((current_kids + clients.active_processes) < max_kids ||
+           max_kids == 0) {
```
HenryMiller1: This seems to change max_kids==0 to mean infinite not 0.

HenryMiller1:
#205 is a better version of this, closing this as redundant.  (thanks for the efforts)

PR #149, opened 2016-04-14
Fix crash in Daemon::answer_client_requests

tombailo:
iceccd was regularly crashing on my PC, and investigation showed the
cause to be using an invalid iterator in Daemon::answer_client_requests.
This function calls create_env_finished for each entry in
native_environments, but create_env_finish will sometimes erase the
entry, leaving the calling function with an invalid iterator.

This commit fixes the crash by moving the erase() up to
answer_client_requests where it can be done safely

### DIFF
--- a/daemon/main.cpp
+++ b/daemon/main.cpp
@@ -532,7 +532,7 @@ struct Daemon {
     int working_loop();
     bool setup_listen_fds();
     void check_cache_size(const string &new_env);
-    void create_env_finished(string env_key);
+    bool create_env_finished(string env_key);
 };
 
 bool Daemon::setup_listen_fds()
@@ -1153,7 +1153,7 @@ bool Daemon::finish_get_native_env(Client *client, string env_key)
     return true;
 }
 
-void Daemon::create_env_finished(string env_key)
+bool Daemon::create_env_finished(string env_key)
 {
     assert(native_environments.count(env_key));
     NativeEnvironment &env = native_environments[env_key];
@@ -1174,8 +1174,7 @@ void Daemon::create_env_finished(string env_key)
                 handle_end(it->second, 121);
             }
         }
-        native_environments.erase(env_key);
-        return;
+        return false;
     }
 
     save_compiler_timestamps(env.gcc_bin_timestamp, env.gpp_bin_timestamp, env.clang_bin_timestamp);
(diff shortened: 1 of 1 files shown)


HenryMiller1:
I'd like to see some tests for this situation if you can figure out how to write them

daemon/main.cpp
```diff
@@ -1884,10 +1884,15 @@ int Daemon::answer_client_requests()
...
+                 it != native_environments.end(); ) {
                 if (it->second.create_env_pipe && FD_ISSET(it->second.create_env_pipe, &listen_set)) {
-                    create_env_finished(it->first);
+                    if(!create_env_finished(it->first))
+                    {
+                        native_environments.erase(it++);
+                        continue;
+                    }
                 }
+                ++it;
```
HenryMiller1: Why did you move this out of the for loop?

HenryMiller1: That is badly worded - why did you move this out of the for command to a line of its own?

johnmiked15: Because `erase()` invalidates the iterator:

```c++
   native_environments.erase(it++);
```
`erase()` is called with the value of the current `it`, but is incremented before the next loop iteration since it uses post-increment. If it's not getting erased, then it needs to be incremented anyway, hence moving it out the for loop. You can't call `erase(it)` within the body of the for loop, because then the `it++` operation that happens after each loop iteration will be using an invalidated `it`.

HenryMiller1: I see, I missed that.  (and had to double check the rules on iterator invalidation for std::map - looks like this is acceptable.

Doesn't this skip the next element in the map if an erase happens?

johnmiked15: No...this is a pretty standard loop that erases elements based on a condition. It looks fine to me.

HenryMiller1: Ahh, the next line is continue...  My mistake.

HenryMiller1:
The change looks good, but I'd like to see some test for this situation so it isn't reintroduced.

tombailo:
Hi all, sorry for the radio silence on this. I'm happy to add some tests for this, but can you point me towards some suitable examples? I can't see any tests for the daemon in the tests/ directory

HenryMiller1:
I'm not sure if the current test approach lets you access details like this.  I just want contributors to think about testing.

Issue #150, opened 2016-04-17
feature request: add an icecc-sheduler option to reject iceccd versions that are too old


mostynb:
1.1rc2 now has split-dwarf support, but it is not safe to use on a possibly mixed version icecc network- pre 1.1rc2 iceccd's don't support split-dwarf debug symbols, and will force clients attempting to use this into local compilation.  Therefore, I think it would be useful to have an icecc-scheduler option to reject such iceccd versions, with a useful message, eg something like "Connection rejected- icecc version X or later is required with this scheduler".

I suppose that an icecc-scheduler command line option to set MIN_PROTOCOL_VERSION would suffice.  Would such a patch be welcome?

johnmiked15:
That's not entirely true. The communication protocol has a version built in and is communicated between the scheduler and various daemons.

I increased the protocol version for -gsplit-dwarf to 35 and made it backward compatible. What this means is that using a new (1.1rc2 or newer) client and daemon locally will let you use the flag when doing local jobs, and will attempt to tell other daemons that are protocol version 35 or newer to do so as well. If it happens to hit a daemon with an older version, the file will still compile, it just won't use -gsplit-dwarf. The returned object file will still work fine, it just won't be split like others would be.

An older local client and daemon will just build everything that has that flag locally as it did before.

That said, an *optional* command line flag or environment variable for the scheduler to enforce a minimum protocol version for connecting daemons is not a terrible idea.

mostynb:
It seems that I misunderstood by debug build logs- when building with -gsplit-dwarf and the job is sent to a version of iceccd that doesn't support this option, ccache sees that the .dwo file is missing and then calls icecc again (after a "Failed; falling back to running the real compiler" log message).  While not as bad as compiling locally, it's could still be a significant performance hit.

Here's an excerpt from my ccache.log for a minimal test case:
[2016-04-17T20:33:19.328294 22056] Executing /usr/bin/icecc /usr/bin/g++ -gsplit-dwarf -c -o hello.o /home/user/.ccache/tmp/hello.stdout.foo.22056.lwEEhG.ii
[2016-04-17T20:33:19.359688 22056] Unlink /home/user/.ccache/8/f/e4076d68053a760b3c7634807a8c60-17491.o.tmp.stdout.foo.22056.RpZYDN
[...]

PR #151, opened 2016-04-17
add -m <ver> flag to icecc-scheduler

mostynb:
This makes it possible to select the minimum supported protocol
version at icecc-scheduler startup time.  Useful for example if
you only want to allow daemons which support -gsplit-dwarf
the debugging option (protocol version 35).

This seems to work, but needs a bit more testing- my version of icemon can't connect when I set the minimum protocol version to 35 (I will try compiling that instead of using ubuntu's package).

### DIFF
--- a/scheduler/scheduler.cpp
+++ b/scheduler/scheduler.cpp
@@ -46,8 +46,8 @@
 #include <algorithm>
 #include <cassert>
 #include <fstream>
-#include <string>
 #include <stdio.h>
+#include <stdlib.h>
 #include <pwd.h>
 #include "../services/comm.h"
 #include "../services/logging.h"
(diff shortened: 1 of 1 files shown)


HenryMiller1:
Where are the tests?  I really do not like merging code changes that don't make an attempt at writing tests.

HenryMiller1:
I think that we want to allow all icemon to use older protocol versions if it wants, or at least require icemon upgrades on a different scheme from clients.

johnmiked15:
I think this change solves your original intent albeit much more simply:

https://gist.github.com/johnmiked15/d6daddcfb7abf78d2aee82402c28ebd5

llunak:
Given that issue #150, that has lead to this, has been closed, I'm going to close this as well. If you think there's still something to be done, see issue #150.

PR #153, opened 2016-05-06
Fix missed inclusion of signal.h

jimklimov:


coolo:
no reason for the configure check - we include signal.h in other places

jimklimov:
I did not find any check for `signal()` in the `configure.ac`, but maybe I did not look too deeply :-)

If you wish, I can remove the line (do not see how it hurts though)... so far there were other non-portability showstoppers that precluded me building this in a Solarish environment with little effort ;(

coolo:
remove everything but the include - signal.h is portable enough

jimklimov:
Not sure I agree fully, but hey - it's your project ;)
Done.

PR #157, opened 2016-05-24
Allow specyfing an alternate remote name in iceccd

ilor:
The extra name passed as --extra-name or via the ICECC_EXTRA_NAME env
var helps solve local jobs not recognized as local if the IP that the
scheduler sees is not the same as the local bound IP in iceccd, for
example when some IP address translation happens between the node and
the scheduler. This is a fix that does not require a protocol update, but does
require the nodes behind NAT to know their real external IP.

### DIFF
--- a/README.md
+++ b/README.md
@@ -539,6 +539,13 @@ problems configuring broadcast. So you might need the -s option for the
 daemon in any case there. If the monitor can't find the scheduler, use
 USE\_SCHEDULER=\<host\> icemon (or send me a patch :)
 
+Note: If there is NAT between a daemon and the scheduler, you may experience
+failures when occasionally jobs bounce back to be compiled on the local daemon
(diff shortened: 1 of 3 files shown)


HenryMiller1:
We need to document this change in the man pages (and probably in Readme.md).  I think I understand what you are getting at from the description here, but I think that a more verbose explanation is needed in the documentation.

ilor:
Sorry for the long delay; can we try to merge this with the doc fixes? We've been using this patch consistently on our build nodes that happen to have a "virtual ip" setup (basically a 1-1 NAT in front of each node, but the node can query the external IP from the virtualization software).

If there are any other issues I can also try to work them out, obviously it would be nice to avoid having to roll a patched iceccd.

ilor:
Thanks!

PR #158, opened 2016-06-09
look for icecc-create-env in the bindir, not the private libdir

mostynb:
icecc-create-env was moved to the bin directory some time ago:
https://github.com/icecc/icecream/commit/5f580f53430aca86fdbf42bb7c2e6cda84be802f

But it seems that icecc --build-native was still looking for icecc-create-env in the private lib directory.  This patch should fix that.

### DIFF
--- a/client/main.cpp
+++ b/client/main.cpp
@@ -183,12 +183,12 @@ static int create_native(char **args)
     vector<char*> argv;
     struct stat st;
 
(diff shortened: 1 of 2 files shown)


HenryMiller1:
Change looks fine, but I'd like to see a test to go with this so that if we move this file again we get notice if we don't make this change.

mostynb:
Added a test.

Issue #159, opened 2016-06-10
-x flag doesn't handle objective-c/c++ files


bgirard:
Pull request incoming. Just doing a test with a large code base first.

PR #160, opened 2016-06-10
Fix -x flag to properly handle objective-c/c++ files

bgirard:
This resolve issue #159 and can now successfully compile the Firefox codebase on a mac using a x86_64 Ubuntu cluster. Seeing more great speed improvements.

### DIFF
--- a/daemon/workit.cpp
+++ b/daemon/workit.cpp
@@ -238,6 +238,7 @@ int work_it(CompileJob &j, unsigned int job_stat[], MsgChannel *client, CompileR
                 argv[i++] = strdup("-Xclang");
                 argv[i++] = strdup(j.workingDirectory().c_str());
             }
+            trace() << "Compiling " << j.inputFile().c_str() << endl;
         }
 
         // HACK: If in / , Clang records DW_AT_name with / prepended .
@@ -289,6 +290,11 @@ int work_it(CompileJob &j, unsigned int job_stat[], MsgChannel *client, CompileR
         // before you add new args, check above for argc
         argv[i] = 0;
         assert(i <= argc);
+
+        for (int currArg = 0; currArg < i; currArg++) {
+          trace() << argv[currArg];
+        }
+        trace() << endl;
 
         close_debug();
 
--- a/daemon/workit.cpp
+++ b/daemon/workit.cpp
@@ -221,7 +221,18 @@ int work_it(CompileJob &j, unsigned int job_stat[], MsgChannel *client, CompileR
         }
 
         argv[i++] = strdup("-x");
-        argv[i++] = strdup((j.language() == CompileJob::Lang_CXX) ? "c++" : "c");
+        if (j.language() == CompileJob::Lang_CXX) {
+          argv[i++] = strdup("c++");
+        } else if (j.language() == CompileJob::Lang_OBJC) {
+          argv[i++] = strdup("objective-c");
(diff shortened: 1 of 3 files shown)


HenryMiller1:
Seems like there should be tests for this.  Even if they have to be optional for those who don't have an objective-c compiler

bgirard:
Sorry but I can't get the test suite to run on any of my machines right now. And I'll be away for 2 weeks.

Here's a test file that can be added that compiles (with a warning but I'd rather not include '@import Foundation;' in the test).

Is it possible to add this test on my behalf?

```
$cat objectivecxx.mm 
@interface ObjC {
    int x;
}
@end
     
@implementation ObjC
- (id)init
{
        return self;
}
@end

int main (int argc, const char * argv[])
{
        return 0;
}
```

bgirard:
FTR on mac I'm guessing the test suite is not portable there. I'm missing setcap/filecap and :
```
if test -x /sbin/setcap; then \
	    sudo /sbin/setcap cap_sys_chroot+ep /Users/bgirard/iceinstall/sbin/iceccd ; \
	elif which filecap >/dev/null 2>/dev/null; then \
	    sudo filecap /Users/bgirard/iceinstall/sbin/iceccd sys_chroot ; \
	else \
	    exit 1 ; \
	fi
```

And with GCC on my Ubuntu machine I get differences in the objdump. Haven't had time to debug why.

bgirard:
Any update on getting this merged in? I have a test case but I'm not able to run the test suite to test it.

daemon/workit.cpp
```diff
@@ -221,7 +221,18 @@ int work_it(CompileJob &j, unsigned int job_stat[], MsgChannel *client, CompileR
...
 
         argv[i++] = strdup("-x");
-        argv[i++] = strdup((j.language() == CompileJob::Lang_CXX) ? "c++" : "c");
+        if (j.language() == CompileJob::Lang_CXX) {
+          argv[i++] = strdup("c++");
+        } else if (j.language() == CompileJob::Lang_OBJC) {
+          argv[i++] = strdup("objective-c");
+        } else if (j.language() == CompileJob::Lang_OBJCXX) {
+          argv[i++] = strdup("objective-c++");
+        } else if (j.language() == CompileJob::Lang_C) {
```
johnmiked15: Can you move the Lang_C conditional to just after CXX? I think the OBJC cases will be used significantly less often.

johnmiked15:
Sorry for the delay. If you can address my inline comment, I'll get this merged in. Tests are nice, but icecream is notoriously difficult to test on such a wide array of possible machine configurations, and your change is harmless enough.

HenryMiller1:
@bgirard can you address the one comment (LANG_C should be earlier in the if/else sequence)?  Seems like a simple fix, but since I don't have a MAC or use objective C I can't really test it.

I think a switch would also work, and might allow the compiler to optimize differently if you prefer.

bgirard:
I'm no longer at Mozilla and I don't have a network that I'm allowed to run icecc on. I pinged my mozilla ex-coworkers who would benefit from this.

llunak:
I've merged the objc(++) patch in, without the other one that's already been done. I have no idea how to add a testcase, but in the worst case this just won't work again, no reason to have this pending. Not that there seems to be much interest in this support one way or another. Thanks for the patch.

bgirard:
Thanks. Mozilla is still using this AFAIK so there's a good chance they might send another PR if it breaks and perhaps a test ;).

PR #165, opened 2016-09-01
recognize versioned gcc/clang binaries

mostynb:
examples of compiler names that we should be able to recognize:
gcc-4.8
g++-4.9
clang-3.6
clang++-3.8

### DIFF
--- a/client/arg.cpp
+++ b/client/arg.cpp
@@ -73,9 +73,15 @@ static bool analyze_program(const char *name, CompileJob &job)
         suffix = compiler_name.substr(compiler_name.size() - 2);
     }
 
-    if ((suffix == "++") || (suffix == "CC")) {
+    // Check for versioned compilers, eg: "gcc-4.8", "clang++-3.6".
+    bool vgcc = compiler_name.find("gcc-") != string::npos;
+    bool vgpp = compiler_name.find("g++-") != string::npos;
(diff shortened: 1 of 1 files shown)


rawicki:
The problem with this change is that icecc runs everything marked as C/C++ job with /usr/bin/gcc or /usr/bin/g++. If under those paths your environment contains something different than specified version you will run into harder to debug problems.

The better solution for this would be - check if the environment contains matching binary and run it remotely, but this is a little bit more work.

mostynb:
I think that's a more general problem- that the user is expected to provide a toolchain tarball that matches the local toolchain.  If you want to attack that problem it might be better done in the daemon, eg cache the hashes of each toolchain that is used.

This patch simply makes it possible to use a icecc with a specific compiler version provided by your linux distribution without resorting to symlink workarounds.  IMO this use-case is worthwhile supporting.

dholbert:
Any likelihood of progress here?

I'm trying to increase `icecc` adoption among my fellow developers at Mozilla (to make good use of a beefy build machine that I recently bought), but several folks have run into this issue -- they're using e.g. `CC=icecc clang-3.6`, and icecream refuses to distribute any builds because it doesn't recognize the compiler.

If @mostynb 's pull request would help with this situation (even as a band-aid), I'm wondering if it'd be possible to take that to improve things.

PR #170, opened 2016-09-23
Translate German logging message to English

dholbert:
This just translates a one-off string of German in the icecc diagnostic output, so that it's in English like all the rest of the icecc logging that I see when I run `iceccd -vvv`.

I checked with some native German speakers and they suggested that "das ist jetzt so" approximately translates to "this is now".  I'm not sure it's a very clear logging message (regardless of German / English), but let's at least make it consistent and not scare people into thinking they're being attacked by German hackers when they see this one-off German string in their logs. :)

### DIFF
--- a/daemon/main.cpp
+++ b/daemon/main.cpp
@@ -1006,7 +1006,7 @@ void Daemon::check_cache_size(const string &new_env)
 
         for (map<string, time_t>::const_iterator it = envs_last_use.begin();
                 it != envs_last_use.end(); ++it) {
-            trace() << "das ist jetzt so: " << it->first << " " << it->second << " " << oldest_time << endl;
+            trace() << "considering cached environment: " << it->first << " " << it->second << " " << oldest_time << endl;
             // ignore recently used envs (they might be in use _right_ now)
             int keep_timeout = 200;
             string native_env_key;


dholbert:
For context, here's what the logging looked like, where I saw this in my own logs:

    [33470] 08:17:12: cleared children
    [33470] 08:17:12: answer_client_requests returned 1
    [33470] 08:17:12: timeout..
    [33470] 08:17:12: accepted 6 127.0.0.1
    [33470] 08:17:12: handle_transfer_env
    [33470] 08:17:12: pid 34414
    [33470] 08:17:12: got pid 34414
    [33470] 08:17:13: handle_transfer_env_done
    [33470] 08:17:13: installed_size: 125242899
    [33470] 08:17:13: installed Darwin16_x86_64/clang_darwin_on_linux size: 125242899 all: 125242899
    [33470] 08:17:13: das ist jetzt so: Darwin16_x86_64/__client 1474643832 1474733833
    [33470] 08:17:13: das ist jetzt so: Darwin16_x86_64/clang_darwin_on_linux 1474643833 1474733833
    [33470] 08:17:13: reannounce_environments
    [33470] 08:17:13: scheduler dead ?!

Note the German (3 & 4 lines from the end) seeming a bit out of place.

HenryMiller1:
Better than it was, but that is still a bad message.  I can't figure out from the code or the log you provided what this message is supported to mean.

daemon/main.cpp
```diff
@@ -1006,7 +1006,7 @@ void Daemon::check_cache_size(const string &new_env)
 
         for (map<string, time_t>::const_iterator it = envs_last_use.begin();
                 it != envs_last_use.end(); ++it) {
-            trace() << "das ist jetzt so: " << it->first << " " << it->second << " " << oldest_time << endl;
+            trace() << "this is now: " << it->first << " " << it->second << " " << oldest_time << endl;
```
HenryMiller1: I think this is supposed to say "considering cached environment"

HenryMiller1:
Please consider what the code is really doing and change the log to make some sense.  Probably that means we need a trace when we enter the function that we are attempting to clear the cache, and then fix this to say considering cache.

dholbert:
It looks like the code is iterating across cached environments, and deciding whether to remove them or not perhaps...? I'm not entirely sure.

Anyway -- I agree that "considering cached environment" seems like a better description here; I don't have any better suggestions than that.

(Really, to the extent that this line of logging is even useful, we might want to go to the source to find out why they really cared to log about this in the first place, and then consider whether it's even useful at this point & if so what it should really say.  But sadly, that's probably not possible, since blame shows that it was added way back in 2007, with commit message "merge in some chunks from make-it-cool that either fix simple things or reduce the diff". https://github.com/icecc/icecream/commit/72dc88c04a8035f67be6a325a650be0b2419ed66#diff-9f80af3e1f09d8bc865a771653b6535bR826 .  So let's just translate it to something reasonable, and let it be improved further as-needed.)

dholbert:
I think I updated the fix, but it seems that created a second commit -- and you probably want a single squashed commit.  (sorry about that, my git/github-fu is not great; when I have to use github I usually rely on its "pencil" tool for one-off minor fixes, but I don't know if it supports squashing/updates)

Let me see if I can push a single squashed commit here.

dholbert:
OK, @HenryMiller1, I think I successfully updated the commit -- could you take another look? Thanks!

dholbert:
Thanks!

PR #174, opened 2016-11-08
Make make Linuxism conditional

da2x:
On FreeBSD, libc provides dl functionality, and there is no separate libdl.

Fixes HEAD’s compilation problems on FreeBSD.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -218,6 +218,9 @@ AC_CHECK_LIB(lzo2, lzo1x_1_compress, LZO_LDADD=-llzo2,
 	AC_MSG_ERROR([Could not find lzo2 library - please install lzo-devel]))
 AC_SUBST(LZO_LDADD)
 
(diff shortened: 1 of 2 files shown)


HenryMiller1:
looks good, but I know nothing about automake...  How much testing did you do on non freebsd?

coolo:
We use autoconf and automake for portability.

coolo:
sorry, but this does not look good. if a -ldl exist or not does not depend on the host os - and it's especially not bound to "linux". That's why autoconf and co were introduced - to avoid hardcoding things and simply test them. A proper patch looks like this one: https://lists.freedesktop.org/archives/wayland-devel/2013-February/007444.html

Aeyoun:
@coolo, I’ll test that approach and change my pull request.

Aeyoun:
Both FreeBSD 10.3 and Fedora Linux 24 seem happy with the new solution.

PR #179, opened 2017-01-23
You can now be able to set custom Scheduler and Daemon Port

GregoireJulien:
You can now launch multiple Scheduler and/or daemon in chroots environement by
setting manually the ports in the icecream configuration file.

I merged manually my old work with this new release, so it might be have some error (I don't hope so...).

To see my changes, in your terminal at the root dir of the sources, search with:

- grep -nre "ahah"
- grep -nre "ohoh"
- grep -nre "ahoh"
- grep -nre "PortCusom"

If the ports are not set, icecream take the default ports.

 I've done thoose modifications to be able to have one big scheduler server, launched in multiples chroots (with different port), and multiples daemon on the same chroots as the scheduler on compilation-nodes. I work in a company who need to have specific chroots env by products. Because the toolchains provided by intel or broadcom or others, makes hard link with files in the system itself...

### DIFF
--- a/client/main.cpp
+++ b/client/main.cpp
@@ -64,6 +64,8 @@
 
 using namespace std;
 
+PortCustomConfig ahah;
+
 extern const char *rs_program_name;
 
 static void dcc_show_usage(void)
@@ -403,7 +405,7 @@ int main(int argc, char **argv)
         }
 
         if (!local_daemon) {
-            local_daemon = Service::createChannel("127.0.0.1", 10245, 0/*timeout*/);
+            local_daemon = Service::createChannel("127.0.0.1", ahah.daemonport, 0/*timeout*/);
         }
     } else {
         local_daemon = Service::createChannel(getenv("ICECC_TEST_SOCKET"));
--- a/scheduler/scheduler.cpp
+++ b/scheduler/scheduler.cpp
@@ -91,6 +91,8 @@
 
 using namespace std;
 
+PortCustomConfig ahoh;
+
 static string pidFilePath;
 
 static map<int, CompileServer *> fd2cs;
@@ -98,7 +100,7 @@ static volatile sig_atomic_t exit_main_loop = false;
 
 time_t starttime;
 time_t last_announce;
-static unsigned int scheduler_port = 8765;
+static unsigned int scheduler_port = ahoh.schedulerport;
 
 // A subset of connected_hosts representing the compiler servers
 static list<CompileServer *> css;
--- a/services/comm.h
+++ b/services/comm.h
@@ -61,6 +61,54 @@
 #define IS_PROTOCOL_34(c) ((c)->protocol >= 34)
 #define IS_PROTOCOL_35(c) ((c)->protocol >= 35)
 
+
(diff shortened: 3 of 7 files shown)


client/main.cpp
```diff
@@ -64,6 +64,8 @@
 
 using namespace std;
 
+PortCustomConfig ahah;
```
HenryMiller1: ahah is a bad name for a variable.  

I'm not comfortable with this as a global either.

HenryMiller1: ahah is a bad name for a variable.

Also I'm not comfortable making this a global.

scheduler/scheduler.cpp
```diff
@@ -91,14 +91,16 @@
 
 using namespace std;
 
+PortCustomConfig ahoh;
```
HenryMiller1: this was ahah above, not ahoh?   

Again this is a bad name.

services/comm.h
```diff
@@ -61,6 +61,54 @@
...
+	unsigned int daemonport;
+	unsigned int schedulerport;
+		
+	char* dp;
+	char* sp;
+	
+	PortCustomConfig()
+	{
+		if(getenv("daemonport") && getenv("schedulerport") != 0)
+		{
```
HenryMiller1: Why do I have to specify both?  I understand for your use case you specify both, but your patch should be reusable for those who only need to change one (ie something else on their network is using scheduler port already, but the saemon port is free)

GregoireJulien: I think it's a mistake made by me.

I my case I need to run multiple daemon in multiple chroot env. We have to plan to use the cgroups classes to be able to manage the cpu power of each chroot env with theere respective daemon.

client/main.cpp
GregoireJulien: yeah I'm not comfortable too, but I'm not a developer. I have few basics knowledges in development. I needed to make this work in a little amount of time :/ 

I test with 3 VM, and it work properly, and compil at the same time without errors.

One VM with two scheduler, and the others VM with two daemon running. It's fine and it run without error, with the custom ports setted. But I agree 100% with you, it's not very clean

HenryMiller1:
This really needs some tests as well: how do we know that some future change won't break this feature?

GregoireJulien:
well, if there is no commit to the port system management, it should not break my modifications. 
It's not a big modification as you can see.

Best regards

HenryMiller1:
While this is a small change, there are numerous problems with it. Painful experience has taught us that adding a feature that works but is not coded well is a negative long term: it makes it harder for the next person to add his change. While I'm reasonably sure this code will work, I can't merge it as it because the code is not well written.

GregoireJulien:
I fully agree with you, I can't write a proper code, because i'm simply not a c++ developper. I'm just a sys Admin with developpement knowledges :-). I'm not take it personnally, I just wanted to share my work, and I hope this may help someone with the same requirements as me  :-)

GregoireJulien:
However, thanks for the replys, I wish you the best for the futur,

Best Regards

PR #189, opened 2017-03-10
Better handle errors extracting environment

HenryMiller1:
I'm investigating errors installing the icecc environment.

Some of my users have found a cases where iceccd is hung (which is to say calling gcc hangs when gcc is icecc), and the user found some processes with "tar | tar | gzip" strung together under the user for icecc (only iceccd is running as this user).  I know this is a little generic, but it is hard to get clear information from these users.  It isn't clear if this fixes their issue but it seems like it could.

### DIFF
--- a/daemon/environment.cpp
+++ b/daemon/environment.cpp
@@ -510,10 +510,11 @@ pid_t start_install_environment(const std::string &basename, const std::string &
     flush_debug();
     pid_t pid = fork();
 
-    if(-1 == pid) {
+    if (pid == -1) {
         log_perror("fork - trying to run tar");
         return 0;
-    } else if (pid) {
+    }
+    if (pid) {
         trace() << "pid " << pid << endl;
         close(fds[0]);
         pipe_to_stdin = fds[1];
@@ -568,7 +569,7 @@ pid_t start_install_environment(const std::string &basename, const std::string &
     argv[5] = 0;
     execv(argv[0], argv);
     log_perror("execv failed");
-    _exit(142);
+    _exit(100);
 }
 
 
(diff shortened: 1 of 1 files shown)


daemon/environment.cpp
```diff
@@ -510,7 +510,10 @@ pid_t start_install_environment(const std::string &basename, const std::string &
     flush_debug();
     pid_t pid = fork();
 
-    if (pid) {
+    if(-1 == pid) {
```
ossilator: missing space.

please don't use yoda conditionals.

daemon/environment.cpp
```diff
@@ -510,7 +510,10 @@ pid_t start_install_environment(const std::string &basename, const std::string &
...
+        log_perror("fork - trying to run tar");
+        return 0;
+    } else if (pid) {
```
ossilator: else unneeded due to return.

daemon/environment.cpp
```diff
@@ -563,7 +566,9 @@ pid_t start_install_environment(const std::string &basename, const std::string &
 
     argv[4] = strdup("-");
     argv[5] = 0;
-    _exit(execv(argv[0], argv));
+    execv(argv[0], argv);
+    log_perror("execv failed");
```
ossilator: it's conventional to pass the file name to perror. the function which failed can be often inferred from that.

daemon/environment.cpp
```diff
@@ -563,7 +566,9 @@ pid_t start_install_environment(const std::string &basename, const std::string &
...
+    _exit(142);
```
ossilator: i don't think you're supposed to recycle the error code.

daemon/environment.cpp
HenryMiller1: That convention (while probably a good idea) is never followed in this entire project.

daemon/environment.cpp
HenryMiller1: I can change it, but ultimately it doesn't matter, anything non-zero is an error are we cannot tell which exit code is from this error vs an error code from tar.

Issue #190, opened 2017-03-10
icecream melts down from long warnings


adam-trhon:
Compiling OpenCV on 4 local cores and 8 remote cores takes us 14 minutes. That's bad, because compiling the same on the local machine only takes 12 minutes. There are big gaps between jobs, it seems that the scheduler does not do its job well.

When we disable warnings (which we don't want to do in production), the distributed build takes 8 minutes. I did some experiments and if I understand it right, icecc cannot distribute other builds when warnings are transferred. And when the transfer takes two or three seconds it hurts the performance very badly.

The only solution I see here is to limit the warning amount. If we have, say, more than 10kB of warnings, we won't read them anyway. How to patch icecream this way? Could we add some env variable to adjust the amount of transferred warnings?

HenryMiller1:
I agree that there is a problem, patches welcome…

That said, I’m not sure why you don’t want to disable warnings.  To my thinking, either you maintain OpenCV and thus want to be running -werror and fixing all warnings, or this is a “third party” library that you don’t maintain and thus are not interested in any warnings anyway.

adam-trhon:
I was completely wrong with the cause estimation. It's not the length, it's the content. This is my sample warning:

    g++ -o f9.o f9.cpp -Wall -c
    In file included from f9.cpp:1:0:
    impl.cpp:6:28: warning: ‘i6’ defined but not used [-Wunused-variable]
     static int APPEND_LINE_NUM(i);
                                ^
    impl.cpp:2:37: note: in definition of macro ‘APPEND_LINE_NUM_3’
     #define APPEND_LINE_NUM_3(ID, LINE) ID##LINE
                                         ^~
    impl.cpp:4:29: note: in expansion of macro ‘APPEND_LINE_NUM_2’
     #define APPEND_LINE_NUM(ID) APPEND_LINE_NUM_2(ID, __LINE__)
                                 ^~~~~~~~~~~~~~~~~
    impl.cpp:6:12: note: in expansion of macro ‘APPEND_LINE_NUM’
     static int APPEND_LINE_NUM(i);
                ^~~~~~~~~~~~~~~

It contains a lot of carets, causing icecc to rebuild the file locally. This is the same warning when I disable the caret workaround using `ICECC_CARET_WORKAROUND`:

    g++ -o f9.o f9.cpp -Wall -c
    In file included from f9.cpp:1:0:
    impl.cpp:6:12: warning: 'i6' defined but not used [-Wunused-variable]

[...]

PR #191, opened 2017-03-15
Disable caret workaround by default

adam-trhon:
Icecream performance suffers greatly when building project with many warnings. This is caused by caret workaround, which rebuilds any file that produces warnings on local node. As a solution, document the workaround and disable it by default.  #Solves #190

### DIFF
--- a/README.md
+++ b/README.md
@@ -19,6 +19,7 @@ Table of Contents
     -   [osc build](#osc-build)
     -   [some compilation node aren't
         used](#some-compilation-node-arent-used)
+    -   [short warnings from GCC](#short-warnings-from-gcc)
 
 -   [Supported platforms](#supported-platforms)
 -   [Using icecream in heterogeneous
@@ -183,6 +184,26 @@ by other nodes, and depending on the scheduler version 12.2 nodes will not compi
 on other nodes either. These incompatible nodes can be identified by having 
 'Linux3_' prefix in the platform). Replace the openSUSE 12.2 package
 with a different one (for example from the devel:tools:build repository).
+
+### short warnings from GCC
+
+GCC4.8+ has -fdiagnostics-show-caret, but when it prints the source code,
+it tries to find the source file on the disk, rather than printing the input
+it got like Clang does. This means that when compiling remotely, it of course
+won't find the source file in the remote chroot, will disable the caret
+silently and print much shorter messages. 
+
+As a workaround, Icecream can recompile locally if there is any stdout/stderr 
+and then print the full message. But it means that files with warnings are 
+compiled twice, once on remote node and one on local node. In projects with many 
+warnigns this has severe impact on performance, so the workaround is disabled
+by default.
+
+To enable it, set
+
+    ICECC_CARET_WORKAROUND=1
+
+in the shell environment where you start the compile jobs.
 
 Supported platforms
 ---------------------------------------------------------------------------------------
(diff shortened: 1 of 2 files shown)


HenryMiller1:
Did you verify test work?  tests/test.sh has code around this workaround.

adam-trhon:
Sorry, I tested it in my environment, but forgot about the tests. I'm on it.

adam-trhon:
@HenryMiller1 what environment should I use for testing? I started with (unchanged) `icecc:master` branch on my Arch (following tests/README), but it failed. Tried on virtual CentOS, failed. Installed virtual OpenSUSE, failed. I used VirtualBox, no docker. My error message:

    Starting icecream.
    Starting icecream successful.

    Running: /usr/bin/g++ -Wall -Werror -gsplit-dwarf -g -c plain.cpp -o /home/adam/Repositories/icecream/tests/results/plain.o
    Files /home/adam/Repositories/icecream/tests/results/plain.o.remote.readelf.txt and /home/adam/Repositories/icecream/tests/results/plain.o.readelf.txt differ
    Output mismatch (/home/adam/Repositories/icecream/tests/results/plain.o.remoteice)
    Makefile:968: recipe for target 'test-run' failed

HenryMiller1:
I’ve never successfully run the tests myself, though I keep trying. Evidence is someone ran them once, and the code hasn’t changed since then.

Can anyone help?

adam-trhon:
@johnmiked15 @mostynb  could you please comment on this? From #161 it appears that you have the tests running

johnmiked15:
What are the contents of those two text files that the tests say are different? This implies that -gsplit-dwarf isn't producing identical object files on your machine. I'll still merge this later, the change is pretty innocuous and seems like the right thing.

adam-trhon:
@johnmiked15 to me it was nothing obvious, as a different rpath, but some "random numbers" there were different. I will start a new issue with the full info. One question before I do so: Do *you* have the tests running somewhere?

I went through the tests quickly and if I understand it right, there are tests that compare error output from local build and remote build. Unless the `ICECC_CARET_WORKAROUND` is set to `1` these tests will fail (at least with gcc).

HenryMiller1:
After reflecting on this more, I believe the current default is correct.

I would like to believe that people are fixing warnings in their code. In this case they want to see the caret.  If you only have a couple warnings, then the performance loss from recompiling locally is insignificant. (though fixing the broken pipe error could make is faster)

Others have brought up warnings in third party libraries (ie boost), but this is not an issue: both gcc and clang have the  -isystem flag which will cause the compiler to not issue warnings for anything in those header files. I will refer you to the documentation on your build system for help. Both compilers have also gotten much better about ensuing that the warnings are right, and have a the ability to enable/disable them individually.

What remains is cases where there are a lot warnings but you are not interested in them. (distribution packages, third party packages, creating a proof of concept for a new platform...) Having to set an environment variable is a good workaround for those cases, and we should document the existence of that variable better so that those who are in a situation of not caring are not slowed down. We should even encourage packagers (ie fedora or debian) to set this by default, but for general users of ice cream the default should be add the caret.

johnmiked15:
It's also worth mentioning that you can disable the caret-infused error messages by passing `-fno-diagnostics-show-caret` as an argument to your compile command in your build. Icecream will notice this switch and not cause a local build if there's anything in stderr when the build job is finished. This lets you selectively turn the workaround on or off as part of your build system. I think this should *also* be documented, and I don't think the workaround should be disabled by default.

@adam-trhon: Yes, I have all of the tests running locally. When I originally implemented `-gsplit-dwarf`, I think I was using Ubuntu 15.10. I just tried them all again with 16.10 and the `-gsplit-dwarf` ones worked fine there too.

HenryMiller1:
#197 replaces this.

Issue #194, opened 2017-03-17
make tests: error code mismatch


HenryMiller1:
When I run tests on ubuntu trusty 14.04 (skipping debug fission tests) they fail

```
Running: /usr/bin/gcc -Wall -Werror -x c++ -c plain -o /home/hm68841/workspace/icecream-upstream/ice-test/tests/results/plain.o
Command successful.

Running: /usr/bin/g++ -c nonexistent.cpp
Local run exit code mismatch (4 vs 1)
make[2]: *** [test-run] Error 2
make[2]: Leaving directory `/home/hm68841/workspace/icecream-upstream/ice-test/tests'
make[1]: *** [test-full] Error 2
make[1]: Leaving directory `/home/hm68841/workspace/icecream-upstream/ice-test/tests'
make: *** [test] Error 2
```

HenryMiller1:
You can see that this is no just my machine:  The same issue exists on

https://travis-ci.org/HenryMiller1/icecream/builds/212256058

johnmiked15:
This makes me think the environment that's getting created doesn't actually match the system. This test works on fine on Ubuntu yakkety. I think the same issue causing this to fail is causing the other failure with the gsplit-dwarf test.

HenryMiller1:
fixed with #196

PR #196, opened 2017-03-20
Do not assume compiler always exits with 1 on error

HenryMiller1:
The compiler exits non-zero on error - this is often one, but cannot be
determined ahead of time


HenryMiller1:
This fixed #194 it turns out when the compiler is run locally on this system it exits with 4 not 1 as the test is expecting.

PR #205, opened 2017-04-12
fix hang on linking when daemon has -m=0

rwcarlsen:
A slightly modified version of #148 that allows leeching daemons with -m=0 while still allowing clients on the leeching machines to run compile/link jobs without hanging.

### DIFF
--- a/daemon/main.cpp
+++ b/daemon/main.cpp
@@ -1210,7 +1210,7 @@ bool Daemon::handle_job_done(Client *cl, JobDoneMsg *m)
 
 void Daemon::handle_old_request()
 {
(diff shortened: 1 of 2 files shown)


HenryMiller1:
It seems like you should be able to write a test for this case.

rwcarlsen:
It would be good to have a test, but I won't have time to work on one in the near term.  For now, we are running a custom version of icecream with this patch.

HenryMiller1:
I'll see what I can do about writing one.  I have some ideas for new features that are likely to break this again so I don't want to merge this without a test.

HenryMiller1:
Replaced by #227

Issue #193, opened 2017-03-17
debug fission tests fail in Ubuntu 14.04


HenryMiller1:
When running tests on ubuntu trust 14.04 they fail as follows.  I have gone back in history and determined that these tests have never passed on trusty since they were introduced.

```
icecc --build-native test successful.

Starting icecream.
Starting icecream successful.

Running: /usr/bin/g++ -Wall -Werror -gsplit-dwarf -g -c plain.cpp -o /home/hm68841/workspace/icecream-upstream/ice-test/tests/results/plain.o
Files /home/hm68841/workspace/icecream-upstream/ice-test/tests/results/plain.o.remote.readelf.txt and /home/hm68841/workspace/icecream-upstream/ice-test/tests/results/plain.o.readelf.txt differ
Output mismatch (/home/hm68841/workspace/icecream-upstream/ice-test/tests/results/plain.o.remoteice)
make[2]: *** [test-run] Error 2
make[2]: Leaving directory `/home/hm68841/workspace/icecream-upstream/ice-test/tests'
make[1]: *** [test-full] Error 2
make[1]: Leaving directory `/home/hm68841/workspace/icecream-upstream/ice-test/tests'
make: *** [test] Error 2
```

The differences look small, but I have not looked into them

```
diff /home/hm68841/workspace/icecream-upstream/ice-test/tests/results/plain.o.remote.readelf.txt /home/hm68841/workspace/icecream-upstream/ice-test/tests/results/plain.o.readelf.txt
4c4
<    Length:        0x73 (32-bit)
---
>    Length:        0xac (32-bit)
15,16c15,16
<     <73>   DW_AT_GNU_pubnames: 1	
<     <73>   DW_AT_GNU_addr_base: 0x0	
---
>     <ac>   DW_AT_GNU_pubnames: 1	
>     <ac>   DW_AT_GNU_addr_base: 0x0	
37c37
<   Size of area in .debug_info section: 288
---
>   Size of area in .debug_info section: 189
40c40
<     109   	f
---
>     a6    	f
47c47
<   Size of area in .debug_info section: 288
---
>   Size of area in .debug_info section: 189
```

HenryMiller1:
You can see that this is not unique to me, https://travis-ci.org/HenryMiller1/icecream/builds/212227718 sees the same thing

johnmiked15:
This test has passed for me since I first committed it. If this isn't passing, I'd like to know what's going on.

What compiler version is trusty using? What version of binutils?

[...]

PR #208, opened 2017-04-17
Fix failing debug fission tests in Ubuntu 14.04

derekgorcz:


HenryMiller1:
Fixes #193

tests/test.sh
```diff
@@ -826,9 +856,10 @@ run_ice "$testdir/includes.o" "remote" 0 $GXX -Wall -Werror -c includes.cpp -o "
 run_ice "$testdir/plain.o" "local" 0 $GXX -Wall -Werror -c plain.cpp -mtune=native -o "$testdir"/plain.o
 run_ice "$testdir/plain.o" "remote" 0 $GCC -Wall -Werror -x c++ -c plain -o "$testdir"/plain.o
 if test -z "$debug_fission_disabled"; then
-    run_ice "" "remote" 1 "split_dwarf" $GXX -gsplit-dwarf -c nonexistent.cpp
+    run_ice "" "remote" "1 4" "split_dwarf" $GXX -gsplit-dwarf -c nonexistent.cpp
```
derekgorcz: g++ (Ubuntu 4.8.4-2ubuntu1~14.04.3) 4.8.4 returns 4 when the input file does not exist.

tests/test.sh
```diff
@@ -252,7 +252,7 @@ run_ice()
     shift
     remote_type="$1"
     shift
-    expected_exit=$1
+    expected_exit=($1)
```
derekgorcz: Added support for multiple possible return values(used on line 859)

tests/test.sh
HenryMiller1: Read the documentation for run_ice - if this is > 128 it will run the command locally to find the error code.  We should use that means  instead of listing all possible exit codes.

This will also remove a bunch of code you added to run_ice

tests/test.sh
```diff
@@ -265,10 +265,15 @@ run_ice()
         shift
     fi
 
-    if [[ $expected_exit -gt 128 ]]; then
-        $@
-        expected_exit=$?
-    fi
+    for (( idx=0; idx<${#expected_exit[@]}; idx++ ))
```
HenryMiller1: This block is not needed - see comment below.

tests/test.sh
HenryMiller1: this change is not needed - see comment below

tests/test.sh
```diff
@@ -364,17 +369,29 @@ run_ice()
         fi
     fi
 
-    remove_debug_info="s/DW_AT_\(GNU_dwo_\(id\|name\)\|comp_dir\|producer\|linkage_name\|name\).*//g"
+    local remove_offset_number='s/<[A-Fa-f0-9]*>/<>/g'
+    local remove_debug_info="s/\(Length\|DW_AT_\(GNU_dwo_\(id\|name\)\|comp_dir\|producer\|linkage_name\|name\)\).*/\1/g"
```
derekgorcz: Added "Length" as an additional field to get removed. Also updated this so that only the value is deleted and the name is left alone. e.g. before it was making the change "<offset>   DW_AT_comp_dir : /some/path  --> <offset>" now it is "<offset>   DW_AT_comp_dir : /some/path  --> <offset> DW_AT_comp_dir"

tests/test.sh
```diff
@@ -364,17 +369,29 @@ run_ice()
         fi
     fi
 
-    remove_debug_info="s/DW_AT_\(GNU_dwo_\(id\|name\)\|comp_dir\|producer\|linkage_name\|name\).*//g"
+    local remove_offset_number='s/<[A-Fa-f0-9]*>/<>/g'
```
derekgorcz: This change removes the offset numbers in the .debug_info section
```
<20>   DW_AT_GNU_dwo_id  : 0xa24190d851ab4b6f   --> <>   DW_AT_GNU_dwo_id  : 0xa24190d851ab4b6f
```

tests/test.sh
```diff
@@ -364,17 +369,29 @@ run_ice()
...
+    local remove_debug_info="s/\(Length\|DW_AT_\(GNU_dwo_\(id\|name\)\|comp_dir\|producer\|linkage_name\|name\)\).*/\1/g"
+    local remove_debug_pubnames='/Offset\s*Name/{n;s/\( *\)[A-Fa-f0-9]*\(\s*.*\)/\1\t\2/}'
```
derekgorcz: g++ 4.8.4 appears to be using a .debug_pubnames section where the newer version uses a .debug_gnu_pubnames section. The gnu version will always have the same "Size of area" value and offset number for the function where the debug_pubnames version will generate different values for there (this happens even when if there are no changes in environment, just running the g++ command twice will produce different output here).

johnmiked15: Use double quotes for consistency on all regular expressions

tests/test.sh
```diff
@@ -646,10 +661,17 @@ debug_test()
...
     # gcc-4.8+ has -grecord-gcc-switches, which makes the .o differ because of the extra flags the daemon adds,
     # this changes DW_AT_producer and also offsets
-    remove_debug_info="s/DW_AT_\(GNU_dwo_\(id\|name\)\|comp_dir\|producer\|linkage_name\|name\).*//g"
-    readelf -wlLiaprmfFoRt "$testdir"/debug-remote.o | sed -e 's/offset: 0x[0-9a-fA-F]*//g' -e 's/[ ]*--param ggc-min-expand.*heapsize\=[0-9]\+//g' -e $remove_debug_info > "$testdir"/readelf-remote.txt
+    local remove_debug_info="s/\(Length\|DW_AT_\(GNU_dwo_\(id\|name\)\|comp_dir\|producer\|linkage_name\|name\)\).*/\1/g"
+    local remove_offset_number='s/<[A-Fa-f0-9]*>/<>/g'
+    local remove_size_of_area='s/\(Size of area in.*section:\)\s*[0-9]*/\1/g'
+    readelf -wlLiaprmfFoRt "$testdir"/debug-remote.o | sed -e 's/offset: 0x[0-9a-fA-F]*//g' \
+        -e 's/[ ]*--param ggc-min-expand.*heapsize\=[0-9]\+//g' \
+        -e $remove_debug_info \
```
johnmiked15: Add quotes around the regular expression variable (and all other occurrences)

PR #217, opened 2017-04-19
Run test binaries from the build directory

### LINKED ISSUE #214: tests run against striped binaries - santizers cannot do backtraces
Icecream tests run against the binaries that are installed.  Install strips these binaries, as a result sanitizers cannot find the symbols for the binaries and thus do not show a useful backtrace.

I was able to work around this by hardcoding in tests/test.sh prefix to be the path to my build directory, and then changing the path to iceccd, icecc, and icecc-scheduler to the appropriate paths in the build directory and get symbols from sanitizer backtraces.  

I think it is enough to make the above production ready, but I'm not sure

HenryMiller1:
Fixes #214

### DIFF
--- a/tests/test.sh
+++ b/tests/test.sh
@@ -9,7 +9,7 @@ builddir=
 
 usage()
 {
-    echo Usage: "$0 <install_prefix> <builddir> [--builddir=dir] [--valgrind[=command]]"
+    echo Usage: "$0 <install_prefix> <testddir> [--builddir=dir] [--valgrind[=command]]"
     exit 3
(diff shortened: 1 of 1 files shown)


HenryMiller1:
@johnmiked15 This seems safe to me and it fixes #214 ( see https://travis-ci.org/HenryMiller1/icecream/jobs/223733304 ) , but this it feels like there might be some other consideration or case that I haven't thought of.

johnmiked15:
Seems fine to me.

PR #227, opened 2017-04-24
Fix hang test

derekgorcz:
This adds a test that verifies code changes in pull request #205


tests/testfunc.cpp
```diff
@@ -0,0 +1,4 @@
+int f()
```
HenryMiller1: Are we no able to reuse the existing files that the other tests use?

tests/test.sh
```diff
@@ -97,6 +97,7 @@ abort_tests()
 start_iceccd()
 {
     name=$1
+    shift
     ICECC_TEST_SOCKET="$testdir"/socket-${name} $valgrind "${iceccd}" 6 -s localhost:8767 -b "$testdir"/envs-${name} -l "$testdir"/${name}.log -N ${name}  -v -v -v "$@" &
```
derekgorcz: seems like this shift was missing, also is the '6' arg a mistake?

tests/testfunc.cpp
derekgorcz: I tried to use plain.cpp instead but it seemed to hang when attempting to compile. Also it seems at a minimum we have to add at-least testmainfunc.cpp in order to test linking functionality.

tests/test.sh
HenryMiller1: seems likely, good catch

HenryMiller1: Can you remove the 6?

HenryMiller1:
Looks like some timing issues: asan is failing.  Maybe we need some waits in those tests or something?

derekgorcz:
I will try and simulate that build locally.

tests/test.sh
```diff
@@ -781,6 +795,8 @@ zero_local_jobs_test()
     remoteice2_pid=$!
     echo $remoteice2_pid > "$testdir"/remoteice2.pid
 
+    wait_for_proc_sleep 10 $localice_pid $remoteice1_pid $remoteice2_pid
+
```
derekgorcz: Once the daemons start, 'ps' reports R+ states for these processes.
```
R    running or runnable (on run queue)
```
```
+    is in the foreground process group
```
The daemons seem to respond once they have reached the 'S' state.
```
S    interruptible sleep (waiting for an event to complete)
```

HenryMiller1:
looks like there are new memory leaks.  (or perhaps you expose existing ones?)  I'm not sure why we are not getting stack traces from the build though, that was working...

derekgorcz:
I configured the build locally based on the yml file and was able to see those initial failures which were fixed based on these changes. However, I did not see these new failures locally.

HenryMiller1:
We are not getting a stack trace because this leak is found in the installed icecream, which has no symbols.  Wherever possible make sure the tests are running against the built versions instead.

tests/Makefile.am
```diff
@@ -2,8 +2,9 @@ test: test-full
 
 test-prepare:
 	if test -x /sbin/setcap; then \
-	    sudo /sbin/setcap cap_sys_chroot+ep ${sbindir}/iceccd ; \
+	    sudo /sbin/setcap cap_sys_chroot+ep ${builddir}/../daemon/iceccd cap_sys_chroot+ep ${sbindir}/iceccd ; \
```
derekgorcz: Adding chroot capability to the daemon binary in the builddir seems to resolve the issue where the test hangs in this new test as well as the debug_test.

tests/test.sh
```diff
@@ -662,9 +677,10 @@ debug_test()
     # debug tests fail when the daemon is not running in the install directory
     # Sanitizers will not give good output on error as a result
     kill_daemon localice
-    ICECC_TEST_SOCKET="$testdir"/socket-localice $valgrind "${prefix}/sbin/iceccd" 6 -s localhost:8767 -b "$testdir"/envs-localice -l "$testdir"/localice.log -N localice  -v -v -v --no-remote -m 2 &
+    ICECC_TEST_SOCKET="$testdir"/socket-localice $valgrind "${prefix}/sbin/iceccd" -s localhost:8767 -b "$testdir"/envs-localice -l "$testdir"/localice.log -N localice  -v -v -v --no-remote -m 2 &
     localice_pid=$!
     echo ${localice_pid} > "$testdir"/${localice}.pid
+    wait_for_proc_sleep 10 $localice_pid
```
derekgorcz: This appears to resolve flaky failures for the debug_test

tests/test.sh
```diff
@@ -758,6 +774,54 @@ debug_test()
...
+zero_local_jobs_test()
+{
+    echo Running zero_local_jobs test.
+
+    reset_logs local "Running zero_local_jobs test"
+    reset_logs remote  "Running zero_local_jobs test"
+
+    kill_daemon localice
+
+    start_iceccd localice --no-remote -m 0
```
derekgorcz: This will now just use the $iceccd version instead of the one in $prefix

derekgorcz:
This last push includes a rebase and squash. The new test is now using the daemons in the build directory but I'm still seeing issues locally when attempting to use the -fsanitize=address and -fno-omit-frame-pointer flags although these failures seem to occur before this new test is run.

derekgorcz:
My guess as to why this new failure is appearing is because the added chroot capability is now causing the CHROOT test to run in addition to the other tests.

derekgorcz:
Issue #235 needs to be resolved before this can be merged.

tests/test.sh
```diff
@@ -758,6 +774,54 @@ debug_test()
...
+
+    wait_for_proc_sleep 10 $localice_pid
+
+    libdir="${testdir}/libs"
+    rm -rf  "${libdir}"
+    mkdir "${libdir}"
+
+    reset_logs remote $GXX -Wall -Werror -c testfunc.cpp -o "${testdir}/testfunc.o"
+    echo Running: $GXX -Wall -Werror -c testfunc.cpp -o "${testdir}/testfunc.o"
+    ICECC_TEST_SOCKET="$testdir"/socket-localice ICECC_TEST_REMOTEBUILD=1 ICECC_PREFERRED_HOST=remoteice1 ICECC_DEBUG=debug ICECC_LOGFILE="$testdir"/icecc.log $valgrind "${icecc}" $GXX -Wall -Werror -c testfunc.cpp -o "${testdir}/testfunc.o"
```
HenryMiller1: you need handle errors here.

HenryMiller1: same for your other compiles and calls to ar below

tests/test.sh
```diff
@@ -758,6 +774,54 @@ debug_test()
...
+    reset_logs local $GXX -Wall -Werror "-L${libdir}" "-ltestlib1" "-ltestlib2" -o "${testdir}/linkedapp"
+    echo Running: $GXX -Wall -Werror "-L${libdir}" "-ltestlib1" "-ltestlib2" -o "${testdir}/linkedapp"
+    ICECC_TEST_SOCKET="$testdir"/socket-localice ICECC_TEST_REMOTEBUILD=1 ICECC_PREFERRED_HOST=localice ICECC_DEBUG=debug ICECC_LOGFILE="$testdir"/icecc.log $valgrind "${icecc}" $GXX -Wall -Werror "-L${libdir}" "-ltestlib1" "-ltestlib2" -o "${testdir}/linkedapp" 2>"$testdir"/stderr.remoteice
+
+    "${testdir}/linkedapp" 2>>"$testdir"/stderr.log
+    app_ret=$?
+    if test ${app_ret} -ne 123; then
+        echo "Error, failed to create a test app by building remotely and linking locally"
+        stop_ice 0
+        exit 2
```
HenryMiller1: use abort_tests not exit

HenryMiller1:
I think we should detect "AddressSanitizer failed to allocate" and skip this test if it happens.

tests/test.sh
HenryMiller1: if you can detect the asan error here and add this to skipped tests I think we are good to go

skipped_tests="$skipped_tests zero_local_jobs_test"

derekgorcz:
There is already one "allow_failures" for the lsan build. Why not just add asan also?

HenryMiller1:
I added lsan because it is part of asan, so running it twice doesn't really make sense anyway.  I really want asan to pass as it is our best assurance that we don't have real memory issues..  As soon as well allow failures you may as well skip the build as nobody will look at it.  (I probably will drop the lsan build soon)

HenryMiller1:
I've had some luck redirecting stderr of iceccd (in start_ieccd) to a file and then greping for this error after the zero test fails.  This isn't ready now though and I won't be working on it this weekend.

derekgorcz:
3e82e03 will skip the zero_local_jobs_test() for the asan build due to the following error.
```
==19173==ERROR: AddressSanitizer failed to allocate 0x10000 (65536) bytes at address 624000050000 (errno: 12)
```
Haven't been able to determine the cause although the clang documentation outlines the following limitations.
- AddressSanitizer uses more stack memory. We have seen up to 3x increase.
- On 64-bit platforms AddressSanitizer maps (but not reserves) 16+ Terabytes of virtual address space. This means that tools like ulimit may not work as usually expected.

PR #229, opened 2017-04-28
check result/errno on all syscalls

ps36094:


client/local.cpp
```diff
@@ -284,36 +284,45 @@ int build_local(CompileJob &job, MsgChannel *local_daemon, struct rusage *used)
         child_pid = fork();
     }
 
-    if (!child_pid) {
+    if (!child_pid || child_pid == -1) {
```
HenryMiller1: if `child_pid == -1` we need to log that. 

I'm not sure running the job is correct in the case where we cannot fork.

ps36094: Ok, I will make that change

client/remote.cpp
```diff
@@ -862,6 +864,11 @@ int build_remote(CompileJob &job, MsgChannel *local_daemon, const Environments &
 
             pid_t pid = fork();
 
+            if (pid == -1) {
+                log_perror("failure of fork");
+                status = -1;
+                }
```
HenryMiller1: this brace is indented too far.

client/remote.cpp
```diff
@@ -915,10 +922,14 @@ int build_remote(CompileJob &job, MsgChannel *local_daemon, const Environments &
                         log_error() << umsgs[i]->hostname << " compiled with exit code " << exit_codes[i]
                                     << " and " << umsgs[0]->hostname << " compiled with exit code "
                                     << exit_codes[0] << " - aborting!\n";
-                        ::unlink(jobs[0].outputFile().c_str());
+                        if (-1 == ::unlink(jobs[0].outputFile().c_str())){
+                            log_perror("unlink failed");
```
HenryMiller1: We should log the filename here too.

daemon/workit.cpp
```diff
@@ -70,6 +70,8 @@ static int death_pipe[2];
 
 extern "C" {
 
+    int ret; /* added */
```
HenryMiller1: the commend doesn't seem to add value.

ps36094: this int ret; has to be removed, and I will do that

PR #230, opened 2017-04-30
log more details when things fail

mostynb:
Here are the beginnings of some error logging improvements- when errors occur on file/directory operations, it's useful to know which file/directory was involved.  Though perhaps we should also include the results of strerror in the output?

### DIFF
--- a/client/remote.cpp
+++ b/client/remote.cpp
@@ -404,7 +404,7 @@ static int build_remote_int(CompileJob &job, UseCSMsg *usecs, MsgChannel *local_
             struct stat buf;
 
             if (stat(version_file.c_str(), &buf)) {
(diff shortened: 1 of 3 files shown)


client/remote.cpp
```diff
@@ -404,7 +404,7 @@ static int build_remote_int(CompileJob &job, UseCSMsg *usecs, MsgChannel *local_
             struct stat buf;
 
             if (stat(version_file.c_str(), &buf)) {
-                log_perror("error stat'ing version file");
```
HenryMiller1: I still want the log_perror.  Can we overload log_perror to return a stream that additional information can be sent to?

mostynb:
The version that landed looks good to me- thanks!

Issue #176, opened 2016-11-24
clang-4.0 tries to read /proc/cpuinfo and failed


kanru:
```
 0:05.57 checking for the target C compiler... /usr/bin/clang
 0:05.69 checking whether the target C compiler can be used... yes
 0:05.69 checking the target C compiler version... 4.0.0
 0:14.61 checking the target C compiler works... no
 0:14.61 DEBUG: Creating `/tmp/conftest.mtiJyc.c` with content:
 0:14.61 DEBUG: |
 0:14.61 DEBUG: | int
 0:14.61 DEBUG: | main(void)
 0:14.61 DEBUG: | {
 0:14.61 DEBUG: |
 0:14.61 DEBUG: |   ;
 0:14.61 DEBUG: |   return 0;
 0:14.61 DEBUG: | }
 0:14.61 DEBUG: Executing: `/usr/bin/icecc /usr/bin/clang -std=gnu99 -c /tmp/conftest.mtiJyc.c`
 0:14.61 DEBUG: The command returned non-zero exit status 105.
 0:14.61 DEBUG: Its error output was:
 0:14.61 DEBUG: | Can't read /proc/cpuinfo: No such file or directory
 0:14.61 DEBUG: | ICECC[6490] 10:54:26: Compiled on 10.247.28.122
 0:14.61 ERROR: Failed compiling a simple C source with the target C compiler
```

coolo:
so how to disable this?

kanru:
I don't know. It looks like it can't be disabled from command line arguments. Maybe I should file a clang bug too.

coolo:
For now I see only one way to fix this: disable remote compilation for clang. There is little value on optimizing for the CPU of the remote host anyway.

kanru:
Ah, please don't disable remote compilation for clang. Since clang 4.0 has not been released yet, I think they will accept a patch that adds flags to control this. Then we can disable it in iceccd.

kito-cheng:
Try --addfile /proc/cpuinfo when use icecc-create-env?

pprkut:
llvm 4.0 was released in the meantime, and the problem is still present.
Creating the environment using "icecc --build-native clang /proc/cpuinfo" however seems to work at least.

AFAICS the change was introduced with https://reviews.llvm.org/D25564 and doesn't even effect compiling, only linking.

HenryMiller1:
This seems like something I want fixed for Release 1.1, but I don't know how to do anything about it. Can someone work on this?

[...]

PR #238, opened 2017-05-05
Document proc/cpuinfo missing error on clang-4.0

HenryMiller1:


HenryMiller1:
This is not a solution to #176 but it will help those with that problem get passed it while we debate the right solution.

PR #239, opened 2017-05-05
Run jobs locally on submitter if possible

RedBeard0531:
This minimizes the latency for single-file edit-compile cycles by only
shipping work over the network when there is more than can be done locally.

I've been running a scheduler with this patch for the past month and it has been working well. I initially wrote it when I noticed that small compiles were taking several seconds longer than they were without icecream.

I don't know this codebase well, so there may be a better way to do this. Ideally we could also give everyone priority on their own box, so no one else can schedule jobs on my box while I have jobs in the queue. But I couldn't figure out how to do that.

### DIFF
--- a/scheduler/scheduler.cpp
+++ b/scheduler/scheduler.cpp
@@ -577,6 +577,12 @@ static CompileServer *pick_server(Job *job)
 
     uint matches = 0;
 
+    if (job->submitter()->maxJobs() > int(job->submitter()->jobList().size())) {
+        // Prefer running locally if the submitter has capacity to minimize
+        // latency when editing and compiling a small number of files.
+        return job->submitter();
+    }
+
     for (list<CompileServer *>::iterator it = css.begin(); it != css.end(); ++it) {
         CompileServer *cs = *it;
 


HenryMiller1:
This is a win for those only building a small number of files, but when you are building a large number of files this is a loss: the local system needs to be kept free to send those files out, not busy compiling others files.  Because of this I do not think this patch should be applied.  I work with a number of cases where we are building a large number of files.

Instead of this patch you should export the environment variable ```ICECC=no``` when you are in that situation.

RedBeard0531:
My goal is that no one needs to think about their build and it just does the "best" thing for them. ie, `build -j400 && run_tests` should Just Work regardless of how many files need to be recompiled. I've been having no trouble keeping our cluster saturated, even when running local builds.

Also, I'm not sure your argument holds. If you are submitting a ton of jobs, won't your machine end up running jobs locally anyway? When you have access to a dedicated compile cluster (rather than just having everyone run iceccd on their own desktop), the best solution may just be to configure your own iceccds with a low -m value.

HenryMiller1:
I don't have a dedicated cluster, but I have a large number of coworkers, so long as we are all don't use low -m we get fast builds on our large project, and the limit is the local CPUs to preprocess and link everything.  Your patch will make things worse for our case.

I understand the desire to do the best thing for everyone, the problem is in this case we have to choose a compromise. You are choosing to optimize for small builds, while I choose to optimize for large builds. Neither choice is inherently wrong, both have pros and cons.

If we are going to do something like this it needs to be a choice on a per-user basis not per machine (in iceccd) or per network (in the scheduler as you did it). This is considerably more complex to implement though. 

We also need tests that it works, which are lacking.

HenryMiller1:
After thinking about this, I think there is a good compromise if we can figure out how to implement it: the client should cache the preprocessed file locally (as well as sending it to the remote).  When preprocessing is complete then the local icecc should ask iceccd how many jobs are in process.  If this is less than the number of CPUs on the system (or configurable number) then build locally as well, otherwise wait for the remote.  That is assume by the time preprocessing completes make has issued all the jobs it can in this build phase so if there are only a few jobs in progress it is probably a small build that should run locally.

Note, if the remote finishes the compile first we should kill the local compile.  (if the local machine is a slow machine and the remote is a fast machine then the remote can build faster)

This requires iceccd to track how many preprocessing jobs are in progress. I would like to do this anyway: when you have a large build cluster the bottleneck is the local CPU doing preprocessing (and linking), and this can take vastly different amounts of times for different files.  When the files need less preprocessing you should run a larger -j than when the files need more - within the same projects this is a different number, so I'd like to have iceccd do more coordination of preprocessing so that we can get better throughput.

llunak:
As said above, this change makes every parallel build worse, so closing this. Commit b5a84d96c4fdc55d6db4c4e3f5a7f44ca48ff2ca should improve the heuristics to make local builds more likely in this case, but as with any heuristic it can't be perfect. I don't think it can be improved that much, maybe some value tweaking (feel free to), but it will be always choosing between preferring local jobs or distributing. The case above with doing both a local and remote build for a seemingly single-job build has, just like my commit, the problem of possibly incorrectly detecting the first job of a large build as a job in a single-job build.

PR #247, opened 2017-05-19
Run tar/gzip at compile priority

derekgorcz:
The priority(nice value) of the local icecc will be used for tar when extracting the environment on a remote server instead of the priority of the iceccd process that is running remotely.

### DIFF
--- a/client/remote.cpp
+++ b/client/remote.cpp
@@ -23,9 +23,11 @@
 
 #include "config.h"
 
-#include <sys/types.h>
+#include <sys/resource.h>
 #include <sys/socket.h>
 #include <sys/stat.h>
+#include <sys/time.h>
+#include <sys/types.h>
 #include <sys/wait.h>
 
 
@@ -410,7 +412,13 @@ static int build_remote_int(CompileJob &job, UseCSMsg *usecs, MsgChannel *local_
                 throw client_error(4, "Error 4 - unable to stat version file");
             }
 
-            EnvTransferMsg msg(job.targetPlatform(), job.environmentVersion());
+            int extract_priority = getpriority(PRIO_PROCESS, 0);
+            if (-1 == extract_priority){
+                log_perror("Unable to determine process priority");
+                extract_priority = 0;
+            }
+
+            EnvTransferMsg msg(job.targetPlatform(), job.environmentVersion(), extract_priority);
 
             if (!cserver->send_msg(msg)) {
                 throw client_error(6, "Error 6 - send environment to remove failed");
--- a/client/remote.cpp
+++ b/client/remote.cpp
@@ -23,11 +23,9 @@
 
 #include "config.h"
 
-#include <sys/resource.h>
+#include <sys/types.h>
 #include <sys/socket.h>
 #include <sys/stat.h>
-#include <sys/time.h>
-#include <sys/types.h>
 #include <sys/wait.h>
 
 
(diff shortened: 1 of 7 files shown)


client/remote.cpp
```diff
@@ -410,7 +412,13 @@ static int build_remote_int(CompileJob &job, UseCSMsg *usecs, MsgChannel *local_
...
             }
 
-            EnvTransferMsg msg(job.targetPlatform(), job.environmentVersion());
+            int extract_priority = getpriority(PRIO_PROCESS, 0);
+            if (-1 == extract_priority){
+                log_perror("Unable to determine process priority");
+                extract_priority = 0;
+            }
+
+            EnvTransferMsg msg(job.targetPlatform(), job.environmentVersion(), extract_priority);
```
derekgorcz: The EnvTransferMsg will now include the 'extract_priority'.

tests/test.sh
```diff
@@ -854,6 +854,74 @@ zero_local_jobs_test()
...
+{
+    echo Running environment_extraction_priority_test test.
+    reset_logs remote "$@"
+
+    mkdir "$testdir/new_test/" &> /dev/null
+
+    test_priorities=(10 12 15)
+    for compile_priority in ${test_priorities[@]}; do
+        kill_daemon remoteice1
+        start_iceccd remoteice1 -p 10246 -m 2
```
derekgorcz: We have to restart the daemon here to clear it's cache which contains the extracted environment from a previous build. If this is not done we won't see tar executing.

tests/test.sh
```diff
@@ -854,6 +854,74 @@ zero_local_jobs_test()
...
+            echo Icecream not ready, aborting.
+            stop_ice 0
+            abort_tests
+        fi
+        sleep 5 # Give the scheduler time to get everything set up
+        flush_logs
+        ICECC_TEST_SOCKET="$testdir"/socket-localice ICECC_TEST_REMOTEBUILD=1 ICECC_PREFERRED_HOST=remoteice1 ICECC_DEBUG=debug ICECC_LOGFILE="$testdir"/icecc.log \
+            nice -n ${compile_priority} $valgrind "${icecc}" $GXX -Wall -Werror -g -c plain.cpp -o "$testdir/new_test/plain.o" 2>"$testdir"/stderr.remoteice &
+        compile_pid=$!
+        for wait_count in $(seq 1 100); do
```
derekgorcz: This loop polls for the existence of the tar process.

client/remote.cpp
HenryMiller1: Also, compile priority is already controlled by the daemon, I think that is the value you use be using.

daemon/environment.cpp
```diff
@@ -589,6 +591,10 @@ pid_t start_install_environment(const std::string &basename, const std::string &
         log_perror("dup2 failed");
     }
 
+    if (-1 == setpriority(PRIO_PROCESS, 0, extract_priority)){
```
HenryMiller1: serve.cpp handle_connection() we use 
'''
int niceval = nice(nice_level);
    if (niceval == -1) {
        log_warning() << "failed to set nice value: " << strerror(errno)
                      << endl;
    }
'''

Seems like we need to be consistent

tests/test.sh
HenryMiller1: This looks like a race condition.  On a loaded machine tar might not start in the expected time, while on a fast machine tar might finish before the test gets here.

I don't know how to test this code... I don't think it is possible to do it with the current style of tests we have.  I think a unit test of some sort is needed. (we currently are not structured to unit test the code in that way)

derekgorcz:
Using a slightly modified version of the removed test, I have been able to verify that the 'tar' process is executed using the 'nice' value which was provided to the remoteice1 daemon using the '--nice' option.
https://travis-ci.org/derekgorcz/icecream/jobs/234653412#L1070
https://github.com/derekgorcz/icecream/commits/test-build

PR #248, opened 2017-05-19
clang related changes

pedronavf:
When integrating clang and the osxcross compiler we noticed several issues:

- ICU uses `-MT "stubdata.d stubdata.o stubdata.ao"`parameter when building, and the client was receiving three parameters `"stubdata.d` `stubdata.o` and `stubdata.ao"`. That was breaking the command line
- osxcross passes `-target x86_64-apple-darwin15`and `-arch x86_64`. The client was interpreting those as file names and was complaining that more than one file was being specified
- the client didn't understand clang's `-c-isystem` and `-cxx-isystem` switches

This pull request addresses all those issues

### DIFF
--- a/client/arg.cpp
+++ b/client/arg.cpp
@@ -404,6 +404,18 @@ bool analyse_argv(const char * const *argv, CompileJob &job, bool icerun, list<s
                        || str_equal("-MG", a)
                        || str_equal("-MP", a)) {
                 args.append(a, Arg_Local);
+            } else if (str_equal("-arch", a)) {
+                args.append(a, Arg_Remote);
+                /* skip next word, being option argument */
+                if (argv[i + 1]) {
+                    args.append(argv[++i], Arg_Remote);
+                }
+            } else if (str_equal("-target", a)) {
+                args.append(a, Arg_Remote);
+                /* skip next word, being option argument */
+                if (argv[i + 1]) {
+                    args.append(argv[++i], Arg_Remote);
+                }
             } else if (str_equal("-fno-color-diagnostics", a)) {
                 explicit_color_diagnostics = true;
                 args.append(a, Arg_Rest);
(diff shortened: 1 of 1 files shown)


HenryMiller1:
Can you write tests for any of this?  Otherwise it looks good.

client/arg.cpp
```diff
@@ -159,7 +159,18 @@ bool analyse_argv(const char * const *argv, CompileJob &job, bool icerun, list<s
                 /* as above but with extra argument */
             } else if (!strcmp(a, "-MT") || !strcmp(a, "-MQ")) {
                 args.append(a, Arg_Local);
-                args.append(argv[++i], Arg_Local);
+                ++i;
+                if (argv[i][0] == '"') {
```
HenryMiller1: I'm wondering if this is best: it seems like several options could have this problem (space in an argument), is there a better way to handle arguments with spaces?

HenryMiller1: I'll call this good enough for now unless someone can suggest a better way.

pedronavf: I agree, I wanted to minimize changes but I think it would be best to check all arguments for quotes, just in case. I think it would be best to have a function that returns the argument value that checks for quotes. Note that we call icecream by wrapping it with a shell script that sets the appropriate environment variables, so it's possible that by escaping them we could get right values passed to icecc. I'll check and revisit this patch.

pedronavf: It turns out that there is no need to handle quotes. I was calling icecc from a shell script that was doing some escaping. I'm reverting the quotes handling patch.

pedronavf:
Sorry, I did something wrong and messed up my branch. I'm submitting a new pull request with some tests and with the quote handling reverted (as the caller should be the one taking care of handling the quotes).

Issue #137, opened 2015-08-11
scheduler should not disconnect clients that are configured to use only that server


HenryMiller1:
We have some icecream users at our physical site that are the lone user on their subnet.  They are able to use ice cream by connecting via scheduler name directly to the scheduler.  However if the configured scheduler decides there is a better scheduler it will disconnect all clients, even those which will never connect to a different one.

The schedulers that are in this situation are well known.  Thus it would be acceptable to have a parameter so the scheduler will not disconnect any client.

PR #249, opened 2017-05-25
Added option to the scheduler to not disconnect clients when a better scheduler is found

ps36094:
…(#137)

### DIFF
--- a/README.md
+++ b/README.md
@@ -108,6 +108,10 @@ cluster after reboot. Consult your distribution's documentation on this. If you
 uses packages provided by your distribution this should be automatic (or a 
 simple configuration change)
 
+How to make scheduler persistent:
+
+By adding an option --scheduler-host for daemon and --persistent-client-connection for scheduler
+,the client connections are not disconnected from the scheduler even there is an availability of better scheduler.
 TroubleShooting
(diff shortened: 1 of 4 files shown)


doc/man-icecc-scheduler.1.xml
```diff
@@ -59,6 +60,11 @@ monitors.</para>
 </varlistentry>
 
 <varlistentry>
+<term><option>-r</option>, <option>--persistent-client-connection</option></term>
+<listitem><para>Client connections are not disconnected from the scheduler if there is a better scheduler available.</para></listitem>
```
HenryMiller1: Client connections are not disconnected from the scheduler EVEN IF...

README.md
```diff
@@ -108,6 +108,10 @@ cluster after reboot. Consult your distribution's documentation on this. If you
 uses packages provided by your distribution this should be automatic (or a 
 simple configuration change)
 
+How to make scheduler persistent:
+
+By adding an option --scheduler-host for daemon and --persistent-client-connection for scheduler
```
HenryMiller1: If seems like this belongs elsewhere in this file, or just in a section not related to "make scheduler persistent"

PR #250, opened 2017-05-26
clang related changes

pedronavf:
When integrating clang and the osxcross compiler we noticed several issues:

- osxcross passes `-target x86_64-apple-darwin15`and `-arch x86_64`. The client was interpreting those as file names and was complaining that more than one file was being specified
- the client didn't understand clang's `-c-isystem` and `-cxx-isystem` switches

This pull request addresses all those issues. Some tests to make sure the new command line switches don't trigger a local build are provided.

### DIFF
--- a/tests/test.sh
+++ b/tests/test.sh
@@ -1053,7 +1053,7 @@ if test -x $CLANGXX; then
     rm "$testdir"/includes.o
     run_ice "" "remote" 0 $CLANGXX -Wall -Werror -cxx-isystem ./ -c includes.cpp -o "$testdir"/includes.o
     rm "$testdir"/includes.o
-    run_ice "" "remote" 0 $CLANGXX -Wall -Werror -target $(uname -m) -c includes.cpp -o "$testdir"/includes.o
+    run_ice "" "remote" 0 $CLANGXX -Wall -Werror -target x86_64-linux-gnu -c includes.cpp -o "$testdir"/includes.o
     rm "$testdir"/includes.o
 
     # test -frewrite-includes usage
--- a/tests/test.sh
+++ b/tests/test.sh
@@ -1051,6 +1051,10 @@ if test -x $CLANGXX; then
     rm "$testdir"/plain.o
     run_ice "" "remote" 0 $CLANGXX -Wall -Werror -c includes.cpp -o "$testdir"/includes.o
     rm "$testdir"/includes.o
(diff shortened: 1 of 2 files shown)


pedronavf:
Ok, I fixed the test problem

tests/test.sh
```diff
@@ -1051,6 +1051,10 @@ if test -x $CLANGXX; then
     rm "$testdir"/plain.o
     run_ice "" "remote" 0 $CLANGXX -Wall -Werror -c includes.cpp -o "$testdir"/includes.o
     rm "$testdir"/includes.o
+    run_ice "" "remote" 0 $CLANGXX -Wall -Werror -cxx-isystem ./ -c includes.cpp -o "$testdir"/includes.o
+    rm "$testdir"/includes.o
+    run_ice "" "remote" 0 $CLANGXX -Wall -Werror -target x86_64-linux-gnu -c includes.cpp -o "$testdir"/includes.o
```
HenryMiller1: will this potentially fail on non x86 platforms if the correct headers are not installed?

pedronavf: when you pass a -target value, clang attempts to find gcc's headers for it by looking at the full expanded triple name, as you can see below (although it's missing a vendor name, hence the double dash after x86_64)

```clang++ -Wall -Werror -c /tmp/a.cpp -target x86_64-linux-gnu --verbose
clang version 4.0.0-1ubuntu1 (tags/RELEASE_400/rc1)
Target: x86_64--linux-gnu
Thread model: posix
InstalledDir: /usr/bin

...

#include <...> search starts here:
 /usr/bin/../lib/gcc/x86_64-linux-gnu/6.3.0/../../../../include/c++/6.3.0
 /usr/bin/../lib/gcc/x86_64-linux-gnu/6.3.0/../../../../include/x86_64-linux-gnu/c++/6.3.0
 /usr/bin/../lib/gcc/x86_64-linux-gnu/6.3.0/../../../../include/c++/6.3.0/backward
 /usr/include/clang/4.0.0/include
 /usr/local/include
 /usr/include/x86_64-linux-gnu
 /usr/include
End of search list.
```
If only ```x86_64``` is passed clang can't find a gcc version for it and it fails to set the include paths. It worked on my system because I have symbolic link as I'm also experimenting with the osxcross project.

For compiling for another target your options are to either install a gcc cross compiler package (clang should be able, in theory, to find it but that behavior is distro dependent), or to manually add -c-isystem/-cxx-isystem entries to point to where the libc/libstdc++ headers are.

Issue #143, opened 2016-03-02
scheduler: stop trying to give jobs to uncooperative clients


ec1oud:
I keep seeing messages like this

ICECC[2983] 09:08:16: no server found behind given hostname 10.-.-.-:10245
ICECC[2983] 09:08:16: got exception 2 (10.-.-.-) 

I don't really understand what it means, but can verify that machine is not listening on that port.

Why can't the scheduler just give up on that client until the problem is corrected?  Ideally with some way to warn the user on that machine too.

HenryMiller1:
This isn't easy - the scheduler cannot just give up on those clients - there might be firewall rules that allow some but not all users to get to that client.  Also, this error is entirely on the client side and doesn't get sent to the scheduler so the scheduler doesn't know that there are bad clients.

Option one the client could have a local blacklist, and when it gets a bad client from the scheduler just ask for another client. I have to check the code, but I think the scheduler will start preferring this bad client as it seems to work so fast...

Option two: the scheduler can have the clients ping each other and report what cannot be reached,thus building a per client blacklist.

Either way we need a way to expire the bad client list.  Otherwise when the problem is fixed the client still won't be used.  Worse, on a network where IPs change often you could end up blacklisting everything.

HenryMiller1:
I see this problem often. I've traced a few to running in a vm but not forwarding the icecc ports to the vm.  The clients can get out but nothing can get back in.

coolo:
the scheduler can check that situation at least - if it can't reach the daemon, it's bad

michaeljs1990:
I am seeing something similar...

```
ICECC[3717] 15:06:48: can't write to /tmp - compiled on 172.30.103.19
ICECC[3717] 15:06:48: got exception 23 (172.30.103.19)
```

I think it might be a litle different but the end result IMO would be the same. the scheduler or client stops trying to give out jobs to this node. I also had restart iceccd on every machine after shutting down the faulty node. Was this the right course of action or would icecc figure it out by itself if I waited longer?

Thanks,

michaeljs1990:
Client side checks seem to be the best way to handle this since like you said large networks can potentially have varying firewall rules. It seem the best way forward would be for the client to keep a list of all bad clients determined by return codes for X time before retrying. On small networks this could be undesirable since one or two failures could mean no icecc at all so a config option to disable this should satisfy all use cases.

AlgorithMan-de:
[...]

PR #251, opened 2017-06-01
If scheduler cannot connect remote machine, send no jobs to it

derekgorcz:


scheduler/compileserver.cpp
```diff
@@ -401,3 +403,52 @@ bool CompileServer::blacklisted(const Job *job, const pair<string, string> &envi
...
+#endif
+    //Time between retries are doubled each failed attempt(seconds)
+    //The delay between retry attempts are capped at 4096s (~1 hr)
+    static const size_t table_size = 12;
+    static const time_t time_offset_table[table_size] = {
+           2,    4,    8,    16,    32,
+          64,  128,  256,   512,  1024,
+        2048, 4096
+    };
+
```
derekgorcz: This table contains the min time between failed connection retries(does not include the timeout (5s) and other blocking code in the main loop). The numbers are generated to double the time each failure but we could get a less drastic increases by using something like e^(-n/a) +1 which would give us values 2 < f(n) < 1 which we can use to multiply against f(n-1) time.

derekgorcz: Also, I am not taking into account the number of servers connected to the scheduler to adjust the retry time.

scheduler/compileserver.cpp
```diff
@@ -401,3 +403,52 @@ bool CompileServer::blacklisted(const Job *job, const pair<string, string> &envi
...
+        delete test_channel;
+        m_inConnAttempt = 0;
+        m_nextConnTime = cur_time + check_back_time;
+        return true;
+    }
+
+    m_nextConnTime = cur_time + time_offset_table[m_inConnAttempt];
+    trace()  << nodeName() << " failed to accept an incoming connection on "
+                << name << ":" << m_remotePort << " attempting again in "
+                << m_nextConnTime - cur_time << " seconds" << endl;
```
derekgorcz: In scheduler.cpp:prune_servers() there is a log message that lets the user know if a server which was accepting connections is no longer doing so and vice versa. This message provides the user with info about retry attempts.

derekgorcz:
Manual testing can be achieved by running the scheduler
```
scheduler/icecc-scheduler -p 8767 -l tests/results/scheduler.log -v -v -v &
```
starting a remote daemon with flag '-p 10246' and blocking the port with
```
sudo /sbin/iptables -A INPUT -p tcp --destination-port 10246 -j DROP
```
The iptables entry can be removed after testing with
```
sudo /sbin/iptables -D INPUT -p tcp --destination-port 10246 -j DROP
```

derekgorcz:
Related to issue #143

scheduler/scheduler.cpp
```diff
@@ -722,7 +723,41 @@ static time_t prune_servers()
...
+            {
+                trace() << "Client (" << (*it)->nodeName() <<
+                    " " << (*it)->name <<
+                    ":" << (*it)->remotePort() <<
+                    ") connected but is not able to accept incoming connections." << endl;
+                //move or copy cs(old) to list css_temp_hold;(copy means that we can still do
+                //disconnection checks and removals from the css list)
+                CompileServer *hold = *it;
+                css_temp_hold.push_back(hold);
+                css.remove(hold);
```
HenryMiller1: seems like css.erase(it) should be better - it is "Linear in the number of elements erased" vs "Linear in container size (comparisons)"

HenryMiller1: Note that we need to get the iterator from find (currently in the if and lost), not it as I wrote.

scheduler/scheduler.cpp
```diff
@@ -722,7 +723,41 @@ static time_t prune_servers()
...
+            }
+            else if (find(css_temp_hold.begin(), css_temp_hold.end(), (*it)) != css_temp_hold.end() && (*it)->incomingConnectionAccepted())
+            {
+                trace() << "Client (" << (*it)->nodeName() <<
+                    " " << (*it)->name <<
+                    ":" << (*it)->remotePort() <<
+                    ") is accepting incoming connections." << endl;
+                CompileServer *restore = *it;
+                css.push_back(restore);
+                css_temp_hold.remove(restore);
```
HenryMiller1: against use erase, we already have the iterator

scheduler/scheduler.cpp
```diff
@@ -105,6 +105,7 @@ static list<CompileServer *> css;
 static list<CompileServer *> monitors;
 static list<CompileServer *> controls;
 static list<string> block_css;
+static list<CompileServer *> css_temp_hold;
```
HenryMiller1: I'm uncomfortable with two lists that we need to maintain: it seems too likely that sometime in the future someone will change something and forget to check both.  I think we should create one class that maintains css, css_temp_hold and css_all. That way there is one source of the information we need and we can just ask for different iterators on it.

HenryMiller1: css_temp_hold isn't a helpful name either.  It is taking me a while to figure out what it does.

scheduler/compileserver.cpp
```diff
@@ -401,3 +403,52 @@ bool CompileServer::blacklisted(const Job *job, const pair<string, string> &envi
...
+}
+
+bool CompileServer::incomingConnectionAccepted()
+{
+#if DEBUG_SCHEDULER > 0
+        trace()  << "Verifying  " << nodeName() << " is accepting incoming connections" << endl;
+#endif
+    //Time between retries are doubled each failed attempt(seconds)
+    //The delay between retry attempts are capped at 4096s (~1 hr)
+    static const size_t table_size = 12;
```
HenryMiller1: can't this be got from sizeof(time_offset_table) as needed?

scheduler/compileserver.cpp
```diff
@@ -401,3 +403,52 @@ bool CompileServer::blacklisted(const Job *job, const pair<string, string> &envi
...
+    static const time_t time_offset_table[table_size] = {
+           2,    4,    8,    16,    32,
+          64,  128,  256,   512,  1024,
+        2048, 4096
+    };
+
+    //On a successful connection, we should still check back every 5min
+    static const time_t check_back_time = 325;
+
+    MsgChannel *test_channel = Service::createChannel(name, remotePort(), 5);
```
HenryMiller1: This is bad - we will block the scheduler for everything else for 5 seconds waiting for a bad client to reply.  The select part of this connect needs to be moved into the main loop of the scheduler.

scheduler/scheduler.cpp
derekgorcz: I didn't want to use the name 'blacklist' because each CompileServer already has a 'm_blacklist' map. Although this new list applies to all CompileServers, I think it may be possible to just have duplicate data and use those existing blacklists which would be less efficient but potentially less confusing.

scheduler/compileserver.cpp
```diff
@@ -152,6 +160,7 @@ bool CompileServer::is_eligible(const Job *job)
            && (m_chrootPossible || job->submitter() == this)
            && load_okay
            && version_okay
+           && m_acceptingInConnection
```
derekgorcz: Instead of using a separate list from css (css_temp_hold) for CompileServers that are not accepting incoming connections, the CompileServer member function is_eligible() will return false.

scheduler/compileserver.cpp
```diff
@@ -401,3 +410,139 @@ bool CompileServer::blacklisted(const Job *job, const pair<string, string> &envi
...
+{
+    return m_inFd;
+}
+
+void CompileServer::setInFd(int fd)
+{
+    m_inFd = fd;
+}
+
+int CompileServer::startInConnectionTest()
```
HenryMiller1: returning an int (fd) from a function of this name is surprising.

scheduler/scheduler.cpp
```diff
@@ -1528,7 +1539,10 @@ static bool handle_end(CompileServer *toremove, Msg *m)
         There might be still clients connected running on the machine on which
          the daemon died.  We expect that the daemon dying makes the client
          disconnect soon too.  */
-        css.remove(toremove);
+        if (find(css.begin(), css.end(), toremove) != css.end())
```
derekgorcz: yes, I missed that

scheduler/scheduler.cpp
```diff
@@ -2245,6 +2276,25 @@ int main(int argc, char *argv[])
                 max_fd--;
             }
         }
+
+        for (list<pair<int, CompileServer *> >::const_iterator it = sock_to_cs_list.begin();
+                it != sock_to_cs_list.end();) {
+            int i = (*it).first;
+            CompileServer *cs = (*it).second;
+            ++it;
```
HenryMiller1: Why is this not in the for?

derekgorcz: will fix

scheduler/scheduler.cpp
```diff
@@ -2245,6 +2273,24 @@ int main(int argc, char *argv[])
...
+            CompileServer *cs = (*it).second;
+
+            if (max_fd && FD_ISSET(i, &read_set) || FD_ISSET(i, &write_set)) {
+                max_fd--;
+
+                cs->inConnectionResponse();
+            }
+            else if (cs->getConnectionInProgress() && cs->getConnectionTimeout() == 0)
+            {
+                close(cs->getInFd());
```
HenryMiller1: This fd belongs to connection - we shouldn't close it here.

scheduler/scheduler.cpp
```diff
@@ -2245,6 +2273,24 @@ int main(int argc, char *argv[])
                 max_fd--;
             }
         }
+
+        for (list<pair<int, CompileServer *> >::const_iterator it = sock_to_cs_list.begin();
+                it != sock_to_cs_list.end(); ++it) {
+            int i = (*it).first;
+            CompileServer *cs = (*it).second;
+
+            if (max_fd && FD_ISSET(i, &read_set) || FD_ISSET(i, &write_set)) {
```
HenryMiller1: It feels like we should just notify CS that select returned, with two bool paramaters: FD_ISSET(i, read_set), FD_ISSET(i, write_set) and let the cs figure out what to do about it.

scheduler/scheduler.cpp
```diff
@@ -723,6 +723,17 @@ static time_t prune_servers()
     }
 
     for (it = css.begin(); it != css.end();) {
+        if ((*it)->startInConnectionTest() != -1)
```
HenryMiller1: I think we should have one call: getNextTimeout and let the cs figure out if it is a new connection or a connection timeout.

scheduler/scheduler.cpp
HenryMiller1: i might be -1, is this legal?  (it is illegal for FD_SET - I can't figure out if it is legal here)

HenryMiller1: I take that back, i cannot be -1...  But this is an implementation detail that would be easy to break in the future.

scheduler/compileserver.cpp
```diff
@@ -401,3 +410,139 @@ bool CompileServer::blacklisted(const Job *job, const pair<string, string> &envi
...
+    remote_addr.sin_port = htons(remotePort());
+    memcpy(&remote_addr.sin_addr.s_addr, host->h_addr_list[0], host->h_length);
+    memset(remote_addr.sin_zero, '\0', sizeof(remote_addr.sin_zero));
+
+    int status = connect(m_inFd, (struct sockaddr *)&remote_addr, sizeof(remote_addr));
+    if ((status < 0) && (errno == EINPROGRESS || errno == EAGAIN))
+    {
+        m_lastConnStartTime=time(0);
+    }
+    else
```
HenryMiller1: isn't there a status==0 case?  I know this is a non-blocking socket but we are not guaranteed that the connection doesn't finish that quick?

scheduler/compileserver.cpp
```diff
@@ -401,3 +410,158 @@ bool CompileServer::blacklisted(const Job *job, const pair<string, string> &envi
...
+        trace()  << nodeName() << " failed to accept an incoming connection on "
+            << name << ":" << m_remotePort << " attempting again in "
+            << m_nextConnTime - time(0) << " seconds" << endl;
+        close(m_inFd);
+        m_inFd = -1;
+    }
+
+}
+
+void CompileServer::inConnectionResponse(int selectRet, fd_set read_set, fd_set write_set)
```
HenryMiller1: We shouldn't be passing a fd_set to this function.  That select is used is an implementation detail of scheduler.cpp.  Just the FD should be passed in.  We might want to switch to poll from select in the future.

HenryMiller1: Actually we know the FD so we don't need it passed in at all.  Likewise we shouldn't need selectRet as this shouldn't be called if select fails.

scheduler/scheduler.cpp
```diff
@@ -2245,6 +2269,11 @@ int main(int argc, char *argv[])
                 max_fd--;
             }
         }
+
+        for (list<CompileServer *>::const_iterator it = cs_in_tsts.begin();
+                it != cs_in_tsts.end(); ++it) {
+            (*it)->inConnectionResponse(max_fd, read_set, write_set);
```
HenryMiller1: We should be checking if the fd is set in either read_set or write_set here.  What we do about it being set belongs to compileServer, but that it is set belongs here.

scheduler/compileserver.cpp
derekgorcz: The select return value will provide us with 0 if there is a timeout on select. We need this information in order to determine that no connection was made as getsockopt(..SO_ERROR..) will return with no errors in this case(getsockopt can let us know about other errors that are not related to a timeout).

derekgorcz: It seems to me like max_fd getting dual-purposed which is a bit confusing. Any thoughts on having max_fd still passed into select but having the select return value stored in a new var, something like 'n_fd_selected'?

PR #255, opened 2017-06-21
icecream - scheduler should not disconnect clients when other server has a different netname

ps36094:


scheduler/scheduler.cpp
```diff
@@ -1816,6 +1823,86 @@ static void trigger_exit(int signum)
...
+                            handle_end(monitors.front(), NULL);
+                        }
+                    }
+                }
+            }
+            else
+            {
+                for(list<CompileServer *>::iterator it = css.begin();it != css.end();++it)
+                {
+                    if ((*it)->netname() == recv_netname)
```
ps36094: we had to add netname to login message so that we know it is OK to disconnect a client if their netname does not match with schedulers netname

services/comm.cpp
```diff
@@ -1859,13 +1859,14 @@ void JobDoneMsg::send_to_channel(MsgChannel *c) const
     *c << flags;
 }
 
-LoginMsg::LoginMsg(unsigned int myport, const std::string &_nodename, const std::string _host_platform)
+LoginMsg::LoginMsg(unsigned int myport, const std::string &_nodename, const std::string &_netname, const std::string _host_platform)
```
ps36094: we had to add netname to login message so that we know it is OK to disconnect a client if their netname does not match with schedulers netname

scheduler/scheduler.cpp
```diff
@@ -1766,16 +1770,19 @@ static int prepare_broadcast_reply(char* buf, const char* netname)
...
     char buf[ schedbuflen ];
     buf[0] = 'I';
     buf[1] = 'C';
     buf[2] = 'E';
     buf[3] = PROTOCOL_VERSION;
     uint64_t tmp_time = starttime;
     memcpy(buf + 4, &tmp_time, sizeof(uint64_t));
+    buf[4 + sizeof(uint64_t)] = length_netname;
+    memcpy(buf + 5 + sizeof(uint64_t), netname, length_netname);
     DiscoverSched::broadcastData(scheduler_port, buf, sizeof(buf));
```
HenryMiller1: We need to send two messages for now, one with the old format (no netname), one with the new one.  Otherwise we break mixed networks.

services/comm.cpp
HenryMiller1: We shouldn't need that.  It isn't the schedulers problem if the client connects to the wrong one, it is a client problem.  

In fact this is a valid and legal case: if someone specifies a specific scheduler to attach to, then matching the netname is not required.

This is also a breaking change which makes this version incompatible with older clients.  I do not want to make such a change.

HenryMiller1: This change would also force a min_protocol_version update.  That is not desirable.

scheduler/scheduler.cpp
```diff
@@ -1766,16 +1770,19 @@ static int prepare_broadcast_reply(char* buf, const char* netname)
     }
 }
 
-static void broadcast_scheduler_version()
+static void broadcast_scheduler_version(const char* netname)
 {
-    const int schedbuflen = 4 + sizeof(uint64_t);
+    char length_netname = sizeof(netname);
```
HenryMiller1: sizeof(const char *) is always 4 or 8.  I think you mean strlen?

HenryMiller1: We need logic to ensure that the total size is <= BROAD_BUFLEN

scheduler/scheduler.cpp
```diff
@@ -1766,16 +1770,19 @@ static int prepare_broadcast_reply(char* buf, const char* netname)
...
+    const int schedbuflen = 5 + sizeof(uint64_t) + length_netname;
     char buf[ schedbuflen ];
```
HenryMiller1: you have to use new/delete here until we have update to C++11.  I want to get this release out with C++98 before forcing the upgrade.  This only compiles now because length_netname is a compiletime constant, but that isn't what you want it to be.

scheduler/scheduler.cpp
```diff
@@ -1766,16 +1770,19 @@ static int prepare_broadcast_reply(char* buf, const char* netname)
...
     buf[0] = 'I';
     buf[1] = 'C';
     buf[2] = 'E';
     buf[3] = PROTOCOL_VERSION;
     uint64_t tmp_time = starttime;
     memcpy(buf + 4, &tmp_time, sizeof(uint64_t));
+    buf[4 + sizeof(uint64_t)] = length_netname;
+    memcpy(buf + 5 + sizeof(uint64_t), netname, length_netname);
```
HenryMiller1: you want strncpy not memcpy though both are equivalent for this purpose.

scheduler/scheduler.cpp
ps36094: When we send a "version 36+" broadcast with the netname, I believe that the old scheduler is still able to use that broadcast data and it will just ignore the additional data that it doesn't use. The call to `recvfrom(... buf, buflen, ...) ` will fill the buffer with only the data that it's expecting (provided by buflen) and any additional data received will be discarded(won't be available for a future call to recvfrom()).

scheduler/scheduler.cpp
```diff
@@ -1816,6 +1823,86 @@ static void trigger_exit(int signum)
...
+       or the same version but was started earlier. */
+    if (!persistent_clients){
+        uint64_t tmp_time;
+        memcpy(&tmp_time, buf + 4, sizeof(uint64_t));
+        time_t other_time = tmp_time;
+        int other_scheduler_protocol = buf[3];
+        string recv_netname;
+        if (other_scheduler_protocol >= 36)
+        {
+            buf[schedbuflen] = '\0';
```
HenryMiller1: We should not modify buf here - future protocols might add more data to the message that we would be clobbering.

scheduler/scheduler.cpp
```diff
@@ -1816,6 +1823,86 @@ static void trigger_exit(int signum)
     signal(signum, trigger_exit);
 }
 
+static void handle_scheduler_announce(char* buf, const char* netname, int schedbuflen, bool persistent_clients, struct sockaddr_in broad_addr)
```
HenryMiller1: the pointer parameters should be const

scheduler/scheduler.cpp
```diff
@@ -1816,6 +1823,86 @@ static void trigger_exit(int signum)
...
+        uint64_t tmp_time;
+        memcpy(&tmp_time, buf + 4, sizeof(uint64_t));
+        time_t other_time = tmp_time;
+        int other_scheduler_protocol = buf[3];
+        string recv_netname;
+        if (other_scheduler_protocol >= 36)
+        {
+            buf[schedbuflen] = '\0';
+            recv_netname = buf + 5 + sizeof(uint64_t);
+            if (recv_netname == netname)
```
HenryMiller1: This cannot possibly work: both of these are pointers which will never be ==.  you want strcmp or better yet make both std::string

scheduler/scheduler.cpp
```diff
@@ -1816,6 +1823,86 @@ static void trigger_exit(int signum)
...
+                            handle_end(css.front(), NULL);
+                        }
+                        while (!monitors.empty())
+                        {
+                            handle_end(monitors.front(), NULL);
+                        }
+                    }
+                }
+            }
+            else
```
HenryMiller1: This entire else should be removed as a don't care condition.

scheduler/scheduler.cpp
```diff
@@ -1816,6 +1821,45 @@ static void trigger_exit(int signum)
...
 
+static void handle_scheduler_announce(const char* buf, const char* netname, bool persistent_clients, struct sockaddr_in broad_addr)
+{
+    /* Another scheduler is announcing it's running, disconnect daemons if it has a better version
+       or the same version but was started earlier. */
+    if (!persistent_clients){
+        uint64_t tmp_time;
+        memcpy(&tmp_time, buf + 4, sizeof(uint64_t));
+        time_t other_time = tmp_time;
+        int other_scheduler_protocol = buf[3];
```
HenryMiller1: this should be a unsigned char not an int.  Also make it const.

scheduler/scheduler.cpp
```diff
@@ -1816,6 +1821,45 @@ static void trigger_exit(int signum)
...
+        string recv_netname;
+        string local_netname = netname;
+        if (other_scheduler_protocol >= 36)
+        {
+            recv_netname = buf + 5 + sizeof(uint64_t);
```
HenryMiller1: declare this variable here, you don't need it outside this scope.

HenryMiller1: This isn't correct, buf is not guaranteed to be null terminated.

HenryMiller1: std::string has a constructor that takes a pointer and a count, I believe that is what you want to use.

HenryMiller1: http://www.cplusplus.com/reference/string/string/string/ - form 5

PR #261, opened 2017-06-28
Added preprocessing load control

derekgorcz:
It appears that the daemon already has process limiting logic in Daemon::handle_old_request() https://github.com/icecc/icecream/blob/master/daemon/main.cpp#L1250. The additional logic added to the client side will inform the daemon of a local build before it calls fork() for a local compile(the fork and exec is done via call_cpp()).

### DIFF
--- a/client/remote.cpp
+++ b/client/remote.cpp
@@ -478,6 +478,23 @@ static int build_remote_int(CompileJob &job, UseCSMsg *usecs, MsgChannel *local_
         }
 
         if (!preproc_file) {
+            local_daemon->send_msg(JobLocalBeginMsg(0, get_absfilename(job.inputFile())));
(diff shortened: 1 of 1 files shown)


HenryMiller1:
Are we handling the case where there is no local daemon running (crashed, or never started)?

HenryMiller1:
I'm not sure if such a case exists as we are already in the remote build, but just checking.

derekgorcz:
It seems that in build_remote() there is already some logic that is blindly dereferencing local_daemon. e.g. local_daemon->send_msg(getcs). In the case that sending fails we are throwing a client_error which is caught in client main function that will return immediately.

derekgorcz:
Also, I believe that if you run the client without a daemon for it to connect to it will just throw and error and halt execution.

Issue #128, opened 2015-04-20
when icecc is from a chroot with a ICECC_VERSION, builds on local machine fail.


HenryMiller1:
If the user is running a build from inside a chroot, connected to a iceccd daemon running on the local system (ie not in the chroot), build that are run on the local system will fail because the environment does not exist.  Builds distributed to other systems will work.

Note, if you want to duplicate this, it is necessary to have the path (unix socket) /var/run/icecc/iceccd.socket mounted from outside the chroot into the chroot so that icecc inside the chroot can communicate with the iceccd deamon running on the local system.

PR #262, opened 2017-07-05
when icecc is from a chroot with a ICECC_VERSION, builds on local machine fail. #128

seerapup:


services/comm.h
```diff
@@ -85,6 +86,8 @@ enum MsgType {
     M_GET_CS,
     // S --> C
     M_USE_CS,  // = 'G'
+    // S --> CS
+    M_NO_CS,
```
HenryMiller1: This reorders the enum and thus breaks compatibility.  You need to add to the end of this enum.

PR #263, opened 2017-07-06
Make clang's -arch and -target Arch_Rest flags

pedronavf:
clang needs to know -arch and -target also when linking, which happens locally. Move those flags to Arg_Rest from Arg_Remote.


HenryMiller1:
Waiting on passing tests.

pedronavf:
ok, I understand what's happening. Arg_Rest arguments are also evaluated when looking for a file name. I'm closing this and opening another pull request once I implement a general way to handle clang's "-argument <value>" parameters.

PR #297, opened 2017-10-08
Fix gentoo make.conf Location

EoD:
The ``make.conf`` location has been moved in end-2012: https://forums.gentoo.org/viewtopic-p-7137202.html

Apparently this README has been written long ago and I am nut sure if the other parts of it are up to date with respect to current Gentoo. Maybe someone else might with more insight can have a look at the other points.

### DIFF
--- a/README.md
+++ b/README.md
@@ -577,7 +577,7 @@ Icecream on gentoo
 -------------------------------------------------------------------------------------
 
 -   It is recommended to remove all processor specific optimizations
-    from the CFLAGS line in /etc/make.conf. On the aKademy cluster it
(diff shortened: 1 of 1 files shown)


HenryMiller1:
I wonder if we should delete all OS specific instructions? I expect that the distributions will keep making changes, and the packagers for them will know how to deal with them better than we do.

EoD:
@HenryMiller1 afaik there is no official gentoo support for icecream (yet?), so it might be a good idea to keep it here in the meanwhile: https://wiki.gentoo.org/index.php?title=Special%3ASearch&search=icecream

Issue #187, opened 2017-02-21
Support coverage builds


P4Cu:
Currently if we detect that there's a coverage flag we're falling back to local build but it should be possible to achieve it over distibuted network if *.gcno/gca files got captured.

I could try to implement that but I would love to here expert opinions first.
- Is it possible with current architecture? (Without hacking the code and making it more messy than it is)
- Which code blocks would need to be modified?

HenryMiller1:
It would be nice to have, but I do not know how much effort it might be.

Dwarf-fission support already brings back two files.  You can probably use https://github.com/icecc/icecream/commit/b3da9eb470cde505a022c0ab2dfeb2e2180235f7 as a guide. I suspect if you expand on that you can make returning more files work.

Best of luck, I haven't actually thought about the problem much.

johnmiked15:
While this is desirable, I'm afraid it would require some significant refactoring and some pretty major design decisions.

When I added debug fission support (to capture .dwo files), I had to hard code the expectation of whether a .dwo file would exist to send back or not. It is not trivial to add more possibilities of output files from a given command line without intimate knowledge of the specific tools involved, which I don't think should be added.

If we were to generalize this concept by making the daemons able to communicate an arbitrary number of output files per compile job, we would need to seriously consider the mechanism for knowing how many to expect, how to find them once the compile job is complete, etc. I do not believe we can do this in a deterministic way, however.

I'm sorry to rain on your parade, but you're probably better off sticking with local builds for this. :-/

P4Cu:
Thanks for an answer. However I'm not giving up that fast ;-)
Two things come to my mind:
1) Sending a package which would be extracted on host.
2) The other thing is rather major but worth consideration. If you think that we should keep things like -ftest-coverage out of ICECC then how about having a plugin system? That's huge task I know.

HenryMiller1:
I think we are just suggesting that you to make sending files back..  I don't think johnmiked15 wants to see more places where we track which files we expect to be produced. I think I agree with him - when I looked at the patch again I see a number of places crying out for a more generic solution, but what that might be I do not know.  We already have binary, the debug file, and you are adding 2 more.  This is enough to call for a generic solution.

[...]

PR #303, opened 2017-10-18
Coverage

svkf:
add extra_file support for code coverage and save_temps
merge split dwarf into that setup
should be backward compatible
added tests for various combinations of the extra files

### DIFF
--- a/client/arg.cpp
+++ b/client/arg.cpp
@@ -728,7 +728,7 @@ bool analyse_argv(const char * const *argv, CompileJob &job, bool icerun, list<s
     job.setOutputFile(ofile);
     if (seen_split_dwarf) {
         // need to do after the output file is set */
-        log_info() << "Seen plitDwarf" << std::endl;
+        log_info() << "Seen SplitDwarf" << std::endl;
         job.setExtraOutputFileEnum(CompileJob::eExFile_dwarfFission);
     }
     if (seen_coverage) {
@@ -738,7 +738,7 @@ bool analyse_argv(const char * const *argv, CompileJob &job, bool icerun, list<s
     }
     if (seen_save_temps) {
         // we really only need this to test if the return protocol will work
-        log_info() << "Save save-temps=cwd" << std::endl;
+        log_info() << "Save save-temps=obj" << std::endl;
         job.setExtraOutputFileEnum(CompileJob::eExFile_st_i);
         job.setExtraOutputFileEnum(CompileJob::eExFile_st_ii);
         job.setExtraOutputFileEnum(CompileJob::eExFile_st_s);
--- a/client/arg.cpp
+++ b/client/arg.cpp
@@ -176,7 +176,6 @@ bool analyse_argv(const char * const *argv, CompileJob &job, bool icerun, list<s
 {
     ArgumentsList args;
     string ofile;
-    string dwofile;
 
 #if CLIENT_DEBUG > 1
     trace() << "scanning arguments" << endl;
@@ -195,6 +194,8 @@ bool analyse_argv(const char * const *argv, CompileJob &job, bool icerun, list<s
     bool seen_mf = false;
     bool seen_md = false;
     bool seen_split_dwarf = false;
+    bool seen_coverage = false;
+    bool seen_save_temps = false;
     // if rewriting includes and precompiling on remote machine, then cpp args are not local
     Argument_Type Arg_Cpp = compiler_only_rewrite_includes(job) ? Arg_Rest : Arg_Local;
 
@@ -328,12 +329,11 @@ bool analyse_argv(const char * const *argv, CompileJob &job, bool icerun, list<s
                 }
             } else if (!strcmp(a, "-S")) {
                 seen_s = true;
-            } else if (!strcmp(a, "-fprofile-arcs")
-                       || !strcmp(a, "-ftest-coverage")
-                       || !strcmp(a, "-frepo")
+            } else if (!strcmp(a, "-frepo")
                        || !strcmp(a, "-fprofile-generate")
                        || !strcmp(a, "-fprofile-use")
                        || !strcmp(a, "-save-temps")
+                       || !strcmp(a, "-save-temps=cwd")
                        || !strcmp(a, "--save-temps")
                        || !strcmp(a, "-fbranch-probabilities")) {
                 log_info() << "compiler will emit profile info (argument " << a << "); building locally" << endl;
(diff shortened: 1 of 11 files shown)


configure.ac
```diff
@@ -55,7 +55,7 @@ if test "$GCC" = yes; then
 -Wshadow -Wpointer-arith $cast_align -Wwrite-strings \
 -Waggregate-return -Wstrict-prototypes -Wmissing-prototypes \
 -Wnested-externs $CFLAGS"
-    CXXFLAGS=" -g -W -Wall -Wpointer-arith $cast_align $wshadow -Wwrite-strings $CXXFLAGS"
+    CXXFLAGS=" -std=gnu++11 -g -W -Wall -Wpointer-arith $cast_align $wshadow -Wwrite-strings $CXXFLAGS"
```
HenryMiller1: This is okay, but I would prefer a minimum level of this, while allowing higher standards if the compiler supports it... (I'm working on cmake which handles this easier than autotools this so don't worry about it)

daemon/main.cpp
```diff
@@ -602,7 +602,7 @@ bool Daemon::setup_listen_fds()
 
     myaddr.sun_family = AF_UNIX;
 
-    mode_t old_umask = -1U;
+    mode_t old_umask = (mode_t)-1U;
```
HenryMiller1: odn't use C style cases - I think you want static_cast here?

daemon/workit.cpp
```diff
@@ -246,21 +242,33 @@ int work_it(CompileJob &j, unsigned int job_stat[], MsgChannel *client, CompileR
         }
 
         bool hasPipe = false;
+        bool hasSaveTemps = false;
 
+        // we could ignore this, except the silly gcc-4.8 thing with tthe caret fails the tests due
```
HenryMiller1: tthe - fix the typo in this comment

daemon/workit.cpp
```diff
@@ -246,21 +242,33 @@ int work_it(CompileJob &j, unsigned int job_stat[], MsgChannel *client, CompileR
...
+        // to having -pipe and -save-temps=obj
+        for (std::list<string>::const_iterator it = list.begin();
```
HenryMiller1: why didn't you use auto here?

HenryMiller1:
It looks good on first glance.  I see some minor style issues. I'll need to look at it a little longer, but thanks for the work.

I'll want @johnmiked15 to look at it too to make sure I'm not missing something, if he is still around

HenryMiller1:
This fixes #187

daemon/main.cpp
svkf: yes, personally I jut hate the extra typing :)

daemon/workit.cpp
svkf: just copied the line below it ...
was sort of last minute, I didn't have this fix in my original since I don't use 4.8 normally but it happened to be on the box I was using

configure.ac
svkf: yeah just the code get so much nicer ... for the for () loops
I can rewrite it the old way ... and remove if you want

svkf:
one issue is the API
the 38 is for all extra_files but then there is an implicit API on the enum/string too
so technically on any new extra_files you would have to raise the number even though the messaging stayed the same
if you don't a new client (-fxxxx) would ask for 38 and if old server (38 but not with new enum) it would fail to get that data
or a old client may use some -fxxx and the remote would look for it (38 with new enum) and send back
but the client wouldn't actually be expecting it since it wouldn't loop through that next one ...

not sure how to handle that .. (just bump API on new enum?) and ensure checked in minVersion?
which is consistent 
could not use enum and tack extra bytes for each enum ... but then that would be more code each change which I don't like

daemon/workit.cpp
svkf: problem was the test failed since it got some stderr and the 4.8 thing
but didn't actually error (even with -Werror), normally for my setup it just kept going ...
gcc: warning: -pipe ignored because -save-temps specified
I was looking at the fix in my setup to just clean it up but figured its so rarely used ... I didn't care (till the test failed)

configure.ac
HenryMiller1: No, I'm saying that if C++14 is available I want to use that instead of C++11. 

I'm not sure how far back we want to support C++ standards, maybe the next release should require C++11, and C++17 for the one after that?

HenryMiller1:
Would it be possible to send to the remote a list of files to send back? Then the version 38 clients are compatible with new file types.  Otherwise just put a comment by the enum that you need to update the protocol version when the enum changes.

svkf:
unfortunately that doesn't help 
(new client, old server)
you send a,b,c  it only can send back a,b
so it would say pass ... but wouldn't actually work the way you want
the scheduler needs to know which guy to pick for what you want 
I could add to the scheduler message GetCSMsg() (in 38) the max enum, then it can use that plus the API#
then its just automatic and no updates needed

(I did something myself adding a custom/API version so that if I added messages on my own I don't pollute the API space on new update from git
(ie naming users, user stats etc.)

client/arg.cpp
```diff
@@ -720,6 +726,23 @@ bool analyse_argv(const char * const *argv, CompileJob &job, bool icerun, list<s
 
     job.setFlags(args);
     job.setOutputFile(ofile);
+    if (seen_split_dwarf) {
+        // need to do after the output file is set */
```
HenryMiller1: remove the c style comment end - this is a C++ comment

tests/test.sh
```diff
@@ -461,6 +504,38 @@ run_ice()
...
+		    abort_tests
+		fi
+	    fi
+	else
+	    echo "No coverage gcno file ($coverage.localice)"
+	    stop_ice 0
+	    abort_tests
+	fi
+    fi
+    # note thiese differ a bit based on the args passed to compiler
```
HenryMiller1: spelling - thiese-> these

services/job.h
```diff
@@ -132,22 +146,71 @@ class CompileJob
...
 
-    void setDwarfFissionEnabled(bool flag)
+    const OutputFiles& outputFiles() const
+    {
+        return m_output_files;
+    }
+
+    void setExtraOutputFile(uint32_t index, const std::string &file)
+    {
+        //printf("setExtraOutputFile[%d] = %s\n", index, file.c_str());
```
HenryMiller1: dead code, please remove, or use the standard logging means to log this

svkf: will do these, just been away

dantje:
Is this PR still maintained? I'd like to look into this and at least make work on master again to experiment with it.

configure.ac
spacemoose: My vote would be c++ 14 for the next release, and 17 after that.   For what it's worth, I've seen several areas of code that could profit from string_view in C++ 17.

deriamis:
I'm going to look into this during the next week or so and see if I can revive it. At this point (years later), I think we should just go for C++17, so I'll see if I can go in that direction - though I would like to keep GNU extensions out of it if I can.

PR #310, opened 2017-11-23
Fixes

llunak:
I haven't contributed to icecream for quite a while, so although I've considered just pushing this directly, I'm submitting these this way, just in case, as I don't know how icecream is currently developed. I may still merge it later if needed.

These are just random small fixes and getting the tests to work for me. There should be nothing complicated or worth discussing[*], I'll submit those separately.

[*] Possible exception being the fake /proc/cpuinfo for clang (issue #176). Clang currently no longer fails, but the error message is still there, it's annoying and it's pointless in this case, as the value is not used anyway. It's unlikely this breaks anything in practice (and even if it later does, that's still better than it being broken now).

### DIFF
--- a/README.md
+++ b/README.md
@@ -202,21 +202,8 @@ There is no known workaround, either disable `-Werror` or fix the code.
 
 ### clang tries to read /proc/cpuinfo and fails
 
-This is a bug in clang 4.0. https://bugs.llvm.org/show_bug.cgi?id=33008 
-It should be fixed in the future, but if you have a broken release you can work around this by
-creating a custom environment and adding /proc/cpuinfo to it.
-
-```
-/usr/lib/icecc/icecc-create-env --clang /usr/bin/clang /usr/lib/icecc/compilerwrapper --addfile /proc/cpuinfo
-```
-
-Do not apply this work around if you do not need it. /proc/cpuinfo is machine specific so and this work 
-around will place wrong information in it. In the case of the bug in clang 4.0 this file is checked for 
-existence but the contents are not actually used, but it is possible future versions of clang/gcc will use
-this file if it exists for something else.
-
-see [Using icecream in heterogeneous environments](#using-icecream-in-heterogeneous-environments) 
-for more information on using icecc-create-env.
+This is a problem of clang 4.0 and newer: https://bugs.llvm.org/show_bug.cgi?id=33008 
+The most recent Icecream version works around this problem.
 
 Supported platforms
 ---------------------------------------------------------------------------------------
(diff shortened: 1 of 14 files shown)


README.md
```diff
@@ -202,21 +202,8 @@ There is no known workaround, either disable `-Werror` or fix the code.
...
-```
-
-Do not apply this work around if you do not need it. /proc/cpuinfo is machine specific so and this work 
-around will place wrong information in it. In the case of the bug in clang 4.0 this file is checked for 
-existence but the contents are not actually used, but it is possible future versions of clang/gcc will use
-this file if it exists for something else.
-
-see [Using icecream in heterogeneous environments](#using-icecream-in-heterogeneous-environments) 
-for more information on using icecc-create-env.
+This is a problem of clang 4.0 and newer: https://bugs.llvm.org/show_bug.cgi?id=33008
```
HenryMiller1: And newer? I asked clang about it and they agreed it was a bug. As I recall they closed it, fixed in 4.1 (which wasn't yet released) and latter.

HenryMiller1:
You should really make your pull requests smaller and independent. There

HenryMiller1:
Large changes like this are hard to review. There seem to be several independent changes in here

llunak:
Well, most of these are minor and I would have normally simply pushed them directly. Do you really want 20 pull requests merely for this? Moreover, at least some of these depend on each other. Also, I noticed you saying somewhere that you are the icecream maintainer now, so what's the status with contributing to icecream now, do you want pull requests for everything, even minor changes? (Note that I have the rights to access directly, I just haven't done anything for several years, so I'm checking.)

WRT the /proc/cpuinfo problem, the current status (clang5) is that it works normally in the absence of /proc/cpuinfo except that clang prints out the warning about cpuinfo missing. And having an extra error message for every compilation in this case is pointless, annoying and it breaks some tests. So it is "fixed" in clang4.0.1+ in the meaning that the build itself no longer breaks, but for icecream's (niche) case it's still sort of broken.

HenryMiller1:
I consider myself maintainer by default, I needed a bug fixed so I did it and then complained enough about it not being merged until someone gave me rights to merge it. I freely admit that I don't understand all of the code. Thus the hacks that seem to work but you realized are wrong....

HenryMiller1:
I generally don't like to merge my own changes. That goes for my day job where I often do understand better than anyone else what the code does. Github would like us to create a contributing.md file, I agree but feel creative one alone is too heavy handed and nobody else sticks with the project long enough to help.

llunak:
I see. Icecream has been around and worked more or less fine for a pretty long time, so it's no wonder people have moved on onto other things after all the time. So it's good that you keep an eye on this. I'll see if I can have an occassional look, but who knows.

PR #311, opened 2017-11-23
Clang tests build fixes

llunak:
These commits make the clang tests build properly/better. This includes some hacks for Travis builds, as it's actually non-trivial to get it to build there. Ubuntu 14.04 apparently ships clang 3.4 as /usr/bin/clang, and icecream in some places still has this path hardcoded. That should indeed get fixed somewhen, but this works for now.

Note that this is only 5 commits, the rest are the same as in my other pull request.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -45,3 +45,4 @@ tests/test-suite.log
 tests/testargs
 tests/testargs.log
 tests/testargs.trs
+tests/test-setup.h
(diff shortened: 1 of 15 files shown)


HenryMiller1:
Please separate out the changes here from the rest. It is hard to review this.

llunak:
The last 5 commits (which is what this pull request is about) are actually all logically the same topic and they depend on each other, it's just that technically the topic "fix clang plugin test" is cleaner to split into more commits that each tackle one problem.

PR #312, opened 2017-11-23
For tests, use installed binaries, not builddir binaries

llunak:
This basically reverts issue #214 , specifically it reverts the logic from ceeff64 and b654aa1. The tests are meant to run from the installation instead from the build, in order to check the installed setup as closely as possible. Moreover, the second commit there basically reverted the first by adding more complexity on top, instead of reverting it as such; and both of those commits had "it fixes sanitizer" as the reason, even though the second is a partial revert of the first one.

There is nothing under normal circumstances that strips the installed binaries, so those commits do not make sense. And as can be seen in https://travis-ci.org/llunak/icecream/jobs/306515182 , sanitizer works just fine.

Note that this is just one commit, the rest are from my other pull requests.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -45,3 +45,4 @@ tests/test-suite.log
 tests/testargs
 tests/testargs.log
 tests/testargs.trs
+tests/test-setup.h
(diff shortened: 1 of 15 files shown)


HenryMiller1:
I agree with the back out. I'm just not sure what fixed things.

PR #317, opened 2017-12-15
don't override clang --target=foo args

mostynb:
1f883606b354dd5fb8b95c8b889ed37aad4e8fbb added some code that handles clang's
"-target foo" flag, but failed to handle the "--target=foo" variant.

I believe this flag is also required when falling back to local compilation.


llunak:
Good catch, it's rather confusing that the two variants are called -target and --target.

> I believe this flag is also required when falling back to local compilation.

No, forced local compiles get all flags.

I'll merge in a modified version of this and a test.

mostynb:
Thanks.

PR #320, opened 2017-12-18
detect gcc more reliably when running tests

mostynb:
If cc is a symlink to gcc, then "cc --version" refers to cc instead of gcc,
then tests are run without using_gcc=1. So let's grep for gcc in the output
of "$TESTCC -v" instead.

### DIFF
--- a/tests/test.sh
+++ b/tests/test.sh
@@ -134,7 +134,7 @@ check_compilers()
         fi
     fi
     using_gcc=
-    if $TESTCC --version | grep gcc >/dev/null; then
+    if $TESTCC -v 2>&1 | grep ^gcc >/dev/null; then
         using_gcc=1
     fi
     using_clang=


mostynb:
Something strange is going on here, I will remove the extra parts of this patch and re-push.

tests/test.sh
```diff
@@ -1476,11 +1476,12 @@ fi
 
 if test -z "$chroot_disabled"; then
     if test -z "$using_gcc"; then
+        # clang jobs don't need the gcc stderr workaround, just spit out the errors.
         run_ice "" "remote" 1 $TESTCXX -c syntaxerror.cpp
-        check_section_log_error icecc "local build forced by remote exception: Error 102 - command needs stdout/stderr workaround, recompiling locally"
```
llunak: This removal is incorrect. It checks that the log message about stderr workaround is NOT present and errors out if it is.

mostynb: Ahh, the difference between the "error" and "message" functions is quite subtle.  Added back.

llunak:
Ok, thank you.

PR #322, opened 2017-12-19
compile fixes for ubuntu 17.10 / GCC 7.2.0

mostynb:
This avoids a couple of warn_unused_result compile errors on Ubuntu 17.10 (which aren't silenced by simply casting to void):

scheduler.cpp: In function ‘int main(int, char**)’:
scheduler.cpp:2019:15: warning: ignoring return value of ‘int daemon(int, int)’, declared with attribute warn_unused_result [-Wunused-result]
         daemon(0, 0);
         ~~~~~~^~~~~~
scheduler.cpp: In function ‘void trigger_exit(int)’:
scheduler.cpp:1801:14: warning: ignoring return value of ‘ssize_t write(int, const void*, size_t)’, declared with attribute warn_unused_result [-Wunused-result]
         write(STDERR_FILENO, msg, strlen( msg ));
         ~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~

### DIFF
--- a/scheduler/scheduler.cpp
+++ b/scheduler/scheduler.cpp
@@ -52,6 +52,7 @@
 #include "../services/comm.h"
 #include "../services/logging.h"
 #include "../services/job.h"
+#include "../services/util.h"
 #include "config.h"
 
 #include "compileserver.h"
@@ -1798,7 +1799,7 @@ static void trigger_exit(int signum)
     } else {
         // hmm, we got killed already. try better
         static const char msg[] = "forced exit.\n";
-        write(STDERR_FILENO, msg, strlen( msg ));
+        ignore_result(write(STDERR_FILENO, msg, strlen( msg )));
         _exit(1);
     }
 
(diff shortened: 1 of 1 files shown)


scheduler/scheduler.cpp
```diff
@@ -1798,7 +1798,11 @@ static void trigger_exit(int signum)
     } else {
         // hmm, we got killed already. try better
         static const char msg[] = "forced exit.\n";
-        write(STDERR_FILENO, msg, strlen( msg ));
+        if (write(STDERR_FILENO, msg, strlen( msg )) == -1) {
+            // Useless, but avoids GCC warn_unused_result error which
+            // can't be silenced by casting to void.
+            log_errno("Error: while writing exit message.", errno);
```
llunak: It's not useless, it may break things. The write() is there presumably because log_errno() might block because of not being async-signal safe, especially in this case when the process doesn't seem to want to exit cleanly. There's ignore_result() somewhere in services/ , see if that helps.

mostynb: Ahh, nice.  Updated.

PR #323, opened 2017-12-20
relax the ASAN test so that it passes on Ubuntu 16.04

mostynb:
For some reason, the ASAN setup on Ubuntu 16.04 lacks proper symbols.
This patch relaxes the ASAN test a little, so that it passes in this
environment.

Ubuntu 17.10 does not require this fix, I'm not sure of the status on
17.04.


llunak:
What exactly is the complete output you get? I'd prefer not to lose the verification of the details about the error.

mostynb:
It seems that I confused the "ERROR" line with the "SUMMARY" line.  The different in the SUMMARY line on Ubuntu 16.04 is that it has "test_fsanitize_function()" instead of "in test_fsanitize_function()".  So we don't need to relax this as much.  Uploading an improved patch...

PR #327, opened 2017-12-26
support @responsefiles via expandargv() from libiberty

dantje:
The library header is wrapped in a namespace to avoid a clash of two
competing basename() functions (std::string and libiberty).

Provides "./configure --without-libiberty" to remove the dependency.

### DIFF
--- a/client/Makefile.am
+++ b/client/Makefile.am
@@ -4,6 +4,7 @@ bin_SCRIPTS = icecc-create-env
 noinst_LIBRARIES = libclient.a
 libclient_a_SOURCES = \
         arg.cpp \
+        argv.c \
         cpp.cpp \
         local.cpp \
         remote.cpp \
--- a/client/argv.c
+++ b/client/argv.c
@@ -1,0 +1,555 @@
+/* Create and destroy argument vectors (argv's)
+   Copyright (C) 1992-2017 Free Software Foundation, Inc.
+   Written by Fred Fish @ Cygnus Support
+
+This file is part of the libiberty library.
+Libiberty is free software; you can redistribute it and/or
+modify it under the terms of the GNU Library General Public
+License as published by the Free Software Foundation; either
+version 2 of the License, or (at your option) any later version.
+
+Libiberty is distributed in the hope that it will be useful,
+but WITHOUT ANY WARRANTY; without even the implied warranty of
+MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+Library General Public License for more details.
+
+You should have received a copy of the GNU Library General Public
+License along with libiberty; see the file COPYING.LIB.  If
+not, write to the Free Software Foundation, Inc., 51 Franklin Street - Fifth Floor,
+Boston, MA 02110-1301, USA.  */
+
+
+/*  Create and destroy argument vectors.  An argument vector is simply an
+    array of string pointers, terminated by a NULL pointer. */
+
+#ifdef HAVE_CONFIG_H
+#include "config.h"
+#endif
+#include "ansidecl.h"
+#include "libiberty.h"
+#include "safe-ctype.h"
+
+/*  Routines imported from standard C runtime libraries. */
+
+#include <stddef.h>
+#include <string.h>
+#include <stdlib.h>
+#include <stdio.h>
+#include <sys/types.h>
+#ifdef HAVE_UNISTD_H
+#include <unistd.h>
+#endif
+#if HAVE_SYS_STAT_H
+#include <sys/stat.h>
+#endif
+
+#ifndef NULL
+#define NULL 0
+#endif
+
+#ifndef EOS
+#define EOS '\0'
+#endif
+
+#define INITIAL_MAXARGC 8	/* Number of args + NULL in initial argv */
+
+
+/*
+
+@deftypefn Extension char** dupargv (char * const *@var{vector})
+
+Duplicate an argument vector.  Simply scans through @var{vector},
+duplicating each argument until the terminating @code{NULL} is found.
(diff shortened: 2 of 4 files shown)


HenryMiller1:
You need to fix the Mac build.

I wonder if this builds on bsd?

dantje:
Looking into it. The libiberty docs themselves propose to ship it alongside your project. Supporting a library that has brittle packaging on multiple platform does not look very inviting.

Ah, and distcc's solution currently is to disable libiberty on osx.

As for the BSDs, I do not have direct access to check that.

Just having a simple c++ implementation of expandargv() (with the same semantics) looks more and more tempting :-)

llunak:
> Just having a simple c++ implementation of expandargv() (with the same semantics) looks more and more tempting :-)

As said, just copy the function from libiberty. It'll save the trouble with configure checks, macosx, bsds and the ugly namespace hack just as with reinventing the wheel.

dantje:
Ok, so I pulled the single required file over into client/argv.c and made some small changes so that we use malloc/strdup instead of xmalloc/xstrdup and friends and removed dead code.

Now Codacy throws some warnings about variables with too generous scopes. These might be fixed.

And it also finds this gem:
```
Condition '!argv_dynamic' is always true
372 if (!argv_dynamic)
```

So if I'm not mistaken expandargv() currently happily duplicates the argv vector for every response file
it encounters on the commandline and leaks a little memory.

llunak:
Yes, it leaks some memory once, that's ok, IIRC libiberty docs even mention that. I think that "always true" report possibly could be avoided by making the variable const, if not, then it can be simply removed. Are you still going to fix the codacy issues? It'd be nice if you did, but I'm ok with merging it as it is too, those can be sorted out later.

dantje:
Fixed the scope issues. I do not like applying a kludge for the "always true" report just to get the number down to zero :-/ It is a bug and once this gets into master, the client will trigger valgrind warnings if 'properly' invoked with a response file:

```
valgrind --leak-check=full client/icecc @responsefile 
==8745== Memcheck, a memory error detector
==8745== Copyright (C) 2002-2015, and GNU GPL'd, by Julian Seward et al.
==8745== Using Valgrind-3.12.0 and LibVEX; rerun with -h for copyright info
==8745== Command: client/icecc @responsefile
==8745== 
ICECC[8745] 14:34:16: connect No such file or directory
ICECC[8745] 14:34:16: connect No such file or directory
ICECC[8745] 14:34:16: connect Connection refused
ICECC[8745] 14:34:16: connect Connection refused
ICECC[8745] 14:34:16: couldn't find any -Wall
ICECC[8745] 14:34:16: could not find -Wall in PATH.
==8745== 
==8745== HEAP SUMMARY:
==8745==     in use at exit: 1,482 bytes in 9 blocks
==8745==   total heap usage: 56 allocs, 47 frees, 97,804 bytes allocated
==8745== 
==8745== 14 bytes in 1 blocks are definitely lost in loss record 3 of 8
==8745==    at 0x4C2DB2F: malloc (in /usr/lib/valgrind/vgpreload_memcheck-amd64-linux.so)
==8745==    by 0x5A91799: strdup (strdup.c:42)
==8745==    by 0x11AD14: dupargv (argv.c:88)
==8745==    by 0x11AD14: expandargv (argv.c:372)
==8745==    by 0x10F1CB: main (main.cpp:226)
==8745== 
==8745== 67 (40 direct, 27 indirect) bytes in 1 blocks are definitely lost in loss record 6 of 8
==8745==    at 0x4C2FD4F: realloc (in /usr/lib/valgrind/vgpreload_memcheck-amd64-linux.so)
==8745==    by 0x11AD9E: expandargv (argv.c:380)
==8745==    by 0x10F1CB: main (main.cpp:226)
==8745== 
==8745== LEAK SUMMARY:
==8745==    definitely lost: 54 bytes in 2 blocks
==8745==    indirectly lost: 27 bytes in 3 blocks
==8745==      possibly lost: 0 bytes in 0 blocks
==8745==    still reachable: 1,401 bytes in 4 blocks
==8745==         suppressed: 0 bytes in 0 blocks
==8745== Reachable blocks (those to which a pointer was found) are not shown.
==8745== To see them, rerun with: --leak-check=full --show-leak-kinds=all
==8745== 
==8745== For counts of detected and suppressed errors, rerun with: -v
==8745== ERROR SUMMARY: 2 errors from 2 contexts (suppressed: 0 from 0)

```

We're in good company though: "valgrind ar @responsefile" also looses 14 bytes :-) So I'd like to work on this for a little while and maybe report the bug back to the gdb/libiberty folks.

llunak:
- Looking at this again, I think the problem with argv_dynamic is that it's supposed to be set after the dupargv() to prevent it being called more than once. There seems to be no need to dupargv() again args that have already been duped.
- Please add a copyright header notice also to the header file.

dantje:
Yes. That seems to be the intention. It needs to make sure that is does not operate on the original argv parameter. But it fails to remember this and just goes on and on to duplicate it again and again for every @responsefile. This code was always broken.

I took the copyright header from another icecream file for the header file.

dantje:
I'm not sure how to tell cppcheck about S_ISDIR. Is there an icecream specific codacy config file somewhere?

llunak:
I have no idea, but don't worry about it, git master has a number of such "issues", so one more is not a blocker.

dantje:
Ok, I'm fine with it. From what I can see, it only leaks what the libiberty docs already describe.

The fix for this leak is to invoke freeargv() at *all* exitpoints in main() iff argv is different from what the init code provided. IMHO there are too many return to make this explicit. So, this calls for some future RAII/scopeguard magic :-)

llunak:
Ok, thanks.

Issue #258, opened 2017-06-27
icecc, run out of memory


FAKERINHEART:
I use the icecream with release version 1.1rc2 for my distributed compilation. However, I find that sometimes the distributed compile server will out of memory while icecc outputs 'stderr: run out of memory'. I have tried decreasing the multiprocessor number of iceccd on each distributed compile server, but this problem still happen when my compilation work is too heavy. It seems that iceccd will use the memory of distributed compile servers unlimitedly.

Is there any method limiting the maximum memory that iceccd uses on on each distributed compile server?

3Q~~~

HenryMiller1:
Can you try the latest master and paste the exact error message you are getting?  We have much better error reporting in master, which might be enough that I can figure out exactly where things failed.

I don't know of a method of limited memory directly. (I'm sure there are kernel tricks you can play but I don't know them) 

Are you sure it is the remotes running out of memory?  Your local machine can run out of memory and that seems more likely to cause  the symptoms you list. Maybe try decreasing the -j count to make.

scottzhang:
I get same issue too. Even I only make -j 1 , I get same output. but sometimes some node success.

[marship@znwg-dhzhjk SMDSPProcessor]$ make -f *.mk
g++  -c  "/data/Projects/smdsolution/SMDSPProcessor/UL33.cpp" -Wall -g -fPIC -std=c++14 -Wno-unknown-pragmas -Wno-reorder -Wno-switch -Wno-deprecated-declarations  -o ./Debug/UL33.cpp.o -I. -I. -I.. -I../include -I../interface -I/usr/local/include/libmongoc-1.0 -I/usr/local/include/libbson-1.0 -I/usr/local/include/mongocxx/v_noabi -I/usr/local/include/bsoncxx/v_noabi -I../devroot/usr/local/include 
ICECC[13404] 20:55:33: connected to /var/run/icecc/iceccd.socket
ICECC[13404] 20:55:33: native /tmp/icecc-envs/native/e6d007abe75c9f8197ecb2d1ab4e9ec9.tar.gz
ICECC[13404] 20:55:33: env: x86_64 '/tmp/icecc-envs/native/e6d007abe75c9f8197ecb2d1ab4e9ec9.tar.gz'
ICECC[13404] 20:55:33: /data/Projects/smdsolution/SMDSPProcessor/UL33.cpp compiled 1 times on x86_64
[...]

PR #328, opened 2017-12-29
log compile job memory limits

mostynb:
This could be useful when debugging compile errors due to OOM, which aren't currently recognised as such.

See also: #258

### DIFF
--- a/daemon/workit.cpp
+++ b/daemon/workit.cpp
@@ -208,12 +208,15 @@ int work_it(CompileJob &j, unsigned int job_stat[], MsgChannel *client, CompileR
 #ifndef SANITIZER_USED
         struct rlimit rlim;
 
-        rlim.rlim_cur = mem_limit * 1024 * 1024;
-        rlim.rlim_max = mem_limit * 1024 * 1024;
+        rlim_t lim = mem_limit * 1024 * 1024;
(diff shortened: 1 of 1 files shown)


tests/test.sh
```diff
@@ -776,6 +776,10 @@ make_test()
     check_log_message remoteice2 "Remote compilation completed with exit code 0"
     check_log_error remoteice2 "Remote compilation aborted with exit code"
     check_log_error remoteice2 "Remote compilation exited with exit code $expected_exit"
+    if [ -z "$SAN_FLAGS" ]; then
+        # The setrlimit call is only made (and logged) in non-sanitizer builds.
+        check_log_message remoteice1 "Compile job memory limit set to "
+    fi
```
llunak: I think this part should be removed, as it actually doesn't do anything in practice (it checks for a presence of a "random" log message, but it doesn't really test or verify anything). Moreover, whether the message is logged or not is not controlled by $SAN_FLAGS, but by a configure check, so it's also incorrect.

mostynb: It indirectly confirms that there was a memory limit set (assuming that SAN_FLAGS is set whenever making sanitizer builds).  But I agree that it's not vital- removed.

PR #330, opened 2017-12-29
add a better description of the memory limit heuristic

mostynb:
I think this heuristic needs tweaking, but I think it's worth understanding the current setup first.

Relates to #258.

### DIFF
--- a/daemon/main.cpp
+++ b/daemon/main.cpp
@@ -835,7 +835,12 @@ bool Daemon::maybe_stats(bool send_ping)
 
 #endif
 
-        // Matz got in the urine that not all CPUs are always feed
+        // Heuristic for the memory limit on compile jobs:
+        //   1) If we have four or more compile slots, allow each job 25% of the available (physical) memory.
+        //   2) If we have three compile slots, allow each job 1/3 of the available (physical) memory.
+        //   3) If we have two compile slots, allow each job 1/2 of the available (physical) memory.
+        //   4) If we have one compile slot, allow it to use all of the available (physical) memory.
+        //   5) Allow each job at least `min_mem_limit` megabytes.
         mem_limit = std::max(int(msg.freeMem / std::min(std::max(max_kids, 1U), 4U)), min_mem_limit);
 
         if (abs(int(msg.load) - current_load) >= 100 || send_ping) {


daemon/main.cpp
```diff
@@ -835,7 +835,12 @@ bool Daemon::maybe_stats(bool send_ping)
...
 #endif
 
-        // Matz got in the urine that not all CPUs are always feed
+        // Heuristic for the memory limit on compile jobs:
+        //   1) If we have four or more compile slots, allow each job 25% of the available (physical) memory.
+        //   2) If we have three compile slots, allow each job 1/3 of the available (physical) memory.
+        //   3) If we have two compile slots, allow each job 1/2 of the available (physical) memory.
+        //   4) If we have one compile slot, allow it to use all of the available (physical) memory.
+        //   5) Allow each job at least `min_mem_limit` megabytes.
         mem_limit = std::max(int(msg.freeMem / std::min(std::max(max_kids, 1U), 4U)), min_mem_limit);
```
llunak: Hmm, is that one line really that cryptic that it needs 6 lines of description? The original comment is indeed rubbish, but this looks like overdoing it way too much in the opposite direction.

mostynb: I expanded this to hopefully get some confirmation that it works as I think it does, before I try to tweak the heuristic (we have started to hit this limit more often after a recent compiler change).  Is it worth shortening this a bit to something like: 

> 1) If we have between one and three compile slots, allow each job 1/n of the available (physical) memory.
> 2) If we have more compile slots, allow each job 1/4 of the available physical memory.
> 3) Allow each job at least `min_mem_limit` megabytes.

Possible changes that I have in mind:
* divide the available memory by the number of unused compile slots (instead of the total number of compile slots, as it seems to be now)
* increase the memory limit if swap is available
* make min_mem_limit a daemon config option
* allow the client to specify a suitable memory limit (or minimum memory limit), but this might be more involved, possibly requires a protocol version bump

llunak:
As far as I can tell, the heuristic works as you described. I personally don't know why it limits it at 4 jobs, my guess would it's from the times when machines didn't have more, and it possibly can be removed.

As for the suggested heuristics, as is often the case with heuristics, it's not as simple:
- divide the available memory by the number of unused compile slots (instead of the total number of compile slots, as it seems to be now) - Tricky. If called at the wrong moment (new jobs starting), mem_limit may be too high.
- increase the memory limit if swap is available - No. Nodes should ideally never notice they're spending their idle time on build jobs for other nodes. If a compilation causes swapping, that's not going to be the case.
- make min_mem_limit a daemon config option - How will anybody know what to set it to and be better at it than the daemon itself?
- allow the client to specify a suitable memory limit (or minimum memory limit), but this might be more involved, possibly requires a protocol version bump - Again, how will they know?

I suggest you should first get some numbers and find out why the current situation is wrong for you before trying to make guesses at how to make it better.

mostynb:
The heuristic was added in 2004 (commit 8bd0b8447c1af5c4f68d098b6d8dc36ffd3e2b48), I guess a high end machine would have been an SMP xeon with hyperthreading, so 4 virtual cores.

> divide the available memory by the number of unused compile slots (instead of the total number of compile slots, as it seems to be now) - Tricky. If called at the wrong moment (new jobs starting), mem_limit may be too high.

This option might require serialization of starting build jobs.

> increase the memory limit if swap is available - No. Nodes should ideally never notice they're spending their idle time on build jobs for other nodes. If a compilation causes swapping, that's not going to be the case.

I have also been thinking about this problem, perhaps swap could count less than ram in the heuristic.  eg apply a scaling factor to swap in the heuristic, perhaps 0.25 * available swap space.  This might be workable because SSD-based swap is much nicer than HDD-based swap.  Over-committing memory works in other scenarios, so I think it's worth at least considering here.  We over-commit memory now when you have more than 4 build slots.

> make min_mem_limit a daemon config option - How will anybody know what to set it to and be better at it than the daemon itself?

In my work environment, we constantly build and rebuild different versions of the same large codebase (chromium).  We start to see patterns over time, and could theoretically give recommendations to the people on the network for how much memory to assign.  This would be arguably better than a hard-coded minimum value chosen in 2004 :)

> allow the client to specify a suitable memory limit (or minimum memory limit), but this might be more involved, possibly requires a protocol version bump - Again, how will they know?

As I mentioned above, it's possible to recognise patterns over time when you're rebuilding the same code.  If it were possible for the client to request a minimum memory limit, then we could add this to our build config and it would work on old and new branches with appropriate values (the general trend is for higher memory requirements as time progresses).

> I suggest you should first get some numbers and find out why the current situation is wrong for you before trying to make guesses at how to make it better.

Of course- that's the plan.

llunak:
I have simply removed the weird comment. I personally consider the line itself reasonably readable. As for the rest, if you still have any interest in working in this, feel free to submit code changes, but for now I'm closing this.

PR #338, opened 2018-01-15
document Arg_* enum values, remove unused Arg_Unspecified

mostynb:
I wonder if it is worthwhile renaming Arg_Rest to something more meaningful, like "Arg_Common" ?

### DIFF
--- a/services/job.h
+++ b/services/job.h
@@ -29,10 +29,9 @@
 #include <sstream>
 
 typedef enum {
-    Arg_Unspecified,
-    Arg_Local,
-    Arg_Remote,
-    Arg_Rest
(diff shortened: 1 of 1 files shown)


mostynb:
I'm guessing that the osx CI error is unrelated.

llunak:
Arg_Remote is not just for flags added by icecc, it can be any flag which makes sense only remotely.

And while Arg_Rest is not exactly clear, I don't find Arg_Common much better (nor I have any better idea).

mostynb:
Tweaked the Arg_Remote comment.

Let's not rename Arg_Rest then (as long as there's a comment for it, I don't mind).

PR #340, opened 2018-01-15
simplify debug_level code

mostynb:
I don't think we need a bit-field for simple verbosity levels, let's
simplify this code a little.


llunak:
Yes, makes sense.

PR #341, opened 2018-01-15
fix typo in icecream manpage

mostynb:
"warning" isn't a supported value for the ICECC_DEBUG environment
variable, should be "warnings".

### DIFF
--- a/client/main.cpp
+++ b/client/main.cpp
@@ -83,7 +83,7 @@ static void dcc_show_usage(void)
         "                              If set to \"disable\", just exec the real compiler, but without\n"
         "                              notifying the daemon and only run one job at a time.\n"
         "   ICECC_VERSION              use a specific icecc environment, see icecc-create-env\n"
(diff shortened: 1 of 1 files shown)


llunak:
Hmm. I guess nobody has used this much, and since all the others are in singular, the proper value should be rather "warning". Could you change this to rather accept both "warning" and also "warnings" for backwards compatibility?

mostynb:
Done.  Do you think it's worth updating the manpage to refer to "warnings" also?

llunak:
The way I see it, since all the others are in singular, the one and only(tm) value should be "warning". The value "warnings" should never be mentioned anywhere (I find it confusing to have both of them listed), it should only be handled for backwards compatibility. Does that make sense?

(Or maybe indeed it should have been only the manpage that needed fixing, I didn't realize --help listed the plural name as well, I originally assumed the "warnings" in the strcmp() was a typo.)

mostynb:
Updated the PR to drop user-visible mention of "warnings".

PR #343, opened 2018-01-16
Clang format ts4

mostynb:
Would something like this be welcome?  Split into two commits to make the .clang-format configuration stand out, and easy to cherry-pick and re-run clang-format if changes land in the meantime.


llunak:
I'm generally against this. There already has been one such change, changing to the KDE style, and this would just change it again to a different style and what would it add besides changing things again and polluting the history?

mostynb:
Can we commit a KDE-style .clang-format file then, and run it on the tree?  The particular format is not important.

llunak:
If I google for kde clang format, it finds a couple of them, but each is different, so I don't know if there's an official one. If you find a usable one, I'm fine with adding it, for the purposes of things like 'git clang-format'.

However I'm still against running it on the whole codebase, I think the code is reasonably well formatted as it is to warrant introducing another unneeded format changes (nor I find it useful to dwell on the precise formatting as long as it's sane and  somewhat consistent).

mostynb:
OK- then I will close this PR and the corresponding issue, and just fix smaller formatting issues as I go.

PR #349, opened 2018-01-18
avoid codacy warning for the S_ISDIR macro

mostynb:
These headers and macros are used unconditionally elsewhere in the tree,
so they should be fine here too.

### DIFF
--- a/client/argv.c
+++ b/client/argv.c
@@ -24,9 +24,7 @@ Boston, MA 02110-1301, USA.  */
 
 #include "argv.h"
 
-#ifdef HAVE_CONFIG_H
 #include "config.h"
-#endif
 
 #include <ctype.h>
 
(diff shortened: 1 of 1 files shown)


client/argv.c
```diff
@@ -37,12 +35,8 @@ Boston, MA 02110-1301, USA.  */
 #include <stdlib.h>
 #include <stdio.h>
 #include <sys/types.h>
-#ifdef HAVE_UNISTD_H
```
HenryMiller1: I'm inclined to suspect this is a bug elsewhere...  But I'm not sure what system doesn't have this header.  configure likes to support odd things from the 1970s which we don't care to support, but it could also be just something some weird bsd doesn't have...

mostynb: The codacy error was only to do with the S_ISDIR macro:
https://www.codacy.com/app/icecc/icecream/pullRequest?prid=1239938
https://github.com/icecc/icecream/pull/343

But this looked like a useless ifdef so I added it to the PR.  I guess this is copied from a third-party codebase that is expected to run on stranger systems than icecc's targets.

PR #351, opened 2018-01-20
icecc-test-env: Add environment test script

JPEWdev:
A helpful environment test script which can be used to find problems
with an environment before sending it over to a remote node

Signed-off-by: Joshua Watt <JPEWhacker@gmail.com>

### DIFF
--- a/client/Makefile.am
+++ b/client/Makefile.am
@@ -1,5 +1,5 @@
 bin_PROGRAMS = icecc
-bin_SCRIPTS = icecc-create-env
+bin_SCRIPTS = icecc-create-env icecc-test-env
 
 noinst_LIBRARIES = libclient.a
 libclient_a_SOURCES = \
--- a/client/icecc-test-env.in
+++ b/client/icecc-test-env.in
@@ -1,0 +1,147 @@
+#! /bin/bash
+#
+# A simple script that can be used to see if an environment was built
+# successfully. Note that passing these test doesn't guarantee the environment
+# will work, but failing them means it certainly won't. Note that this script
+# may need to be executed with sudo if the current user doesn't have chroot
+# permissions
+#
+# This program always exits with an error code of 2 so that it can be
+# distinguished from a sudo error (with an exit code of 1)
+#
+# Copyright (C) 2018 Joshua Watt
+#
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation; either version 2 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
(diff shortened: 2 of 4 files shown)


llunak:
Can you please add usage of this to tests/test.sh ? Function buildnativetest() looks like a good place.

JPEWdev:
Yes, I wasn't sure exactly how to handle an automated test case. Is it OK for it stop the test to prompt for the sudo password, or should it use -f flag to use fakechroot + fakeroot?

JPEWdev:
I realize the test script was already calling sudo, so its probably wasn't a concern that icecc-create-env also did so.

Done

JPEWdev:
I tracked down the asan an lsan errors in Travis. The compilerwrapper is compiled with "-fsanitizer=..." from the exported CXXFLAGS, however the address sanitizer for clang appears to try and read /proc/self/maps, which doesn't exist in the chroot. This causes my test script to fail (in this build), and the build to fail later (in other builds). There doesn't appear to be any way to turn off the sanitizers at runtime, or at least you can't turn them off before it tries to access /proc/self/maps.

It only affects the toolchain when clang is used to compile compilerwrapper, and when TESTCC=clang (which causes compilerwrapper to be included in the environment).

I *think* the solution would be to add -fno-sanitize=all to the compile command line for compilerwrapper, but it would have to come *after* the user CXXFLAGS, which automake doesn't seem willing to do (or at least I don't know how to make automake do it).

llunak:
The test script itself does not call sudo, it's the optional Makefile preparation. Calling test.sh alone should work fine even without sudo.

Can you make the usage in test.sh use sudo with -n? Something like (untested):

retcode=
sudo -n test-env $whatever
retcode=$?
if $retcode -eq 1; then
  if ! command -v fakeroot >/dev/null; then
    echo Cannot explicitly verify environment, use sudo or fakeroot.
    skipped_tests="$skipped_tests testenv"
  else
    test-env --fakeroot $whatever
    retcode=$?
  fi
fi
if test -n "$retcode" -a $retcode -ne 0; then
  # test failed
fi

llunak:
As for compilerwrapper, a simple and reasonable solution would be not bother with it. It's for backwards compatibility with nodes that are several years old by now and if everything else works, presumably a simple wrapper will keep working as well.

If you want to compile it without ASAN, see services/Makefile.am , it sets per-target CXXFLAGS, that should work. In that case you'll need to add a configure check for whether the flag actually exists.

JPEWdev:
The per-target CXXFLAGS are applied by automake before the user supplied CXXFLAGS environment variable, so that doesn't help unfortunately (I need to override the -fsanitize= options the user has provided). I think I'll take the easy route and not test the compiler wrapper in clang since they appear to be deprecated.

llunak:
Ok. Thanks.

PR #357, opened 2018-01-24
[codacy] fix printf formatting issue in tempfile.c

mostynb:
%lx in format string (no. 3) requires 'unsigned long' but the argument type is 'unsigned long long'.

### DIFF
--- a/services/tempfile.c
+++ b/services/tempfile.c
@@ -88,10 +88,10 @@ int dcc_make_tmpnam(const char *prefix, const char *suffix, char **name_ret, int
 #endif
 
     do {
-        if (snprintf(tmpname, tmpname_length, "%s/%s_%08lx%s",
+        if (snprintf(tmpname, tmpname_length, "%s/%s_%08lu%s",
                      (relative ? _PATH_TMP + 1 : _PATH_TMP),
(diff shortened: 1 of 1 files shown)


services/tempfile.c
```diff
@@ -88,10 +88,10 @@ int dcc_make_tmpnam(const char *prefix, const char *suffix, char **name_ret, int
 #endif
 
     do {
-        if (snprintf(tmpname, tmpname_length, "%s/%s_%08lx%s",
+        if (snprintf(tmpname, tmpname_length, "%s/%s_%08lu%s",
                      (relative ? _PATH_TMP + 1 : _PATH_TMP),
                      prefix,
-                     random_bits & 0xffffffffUL,
+                     random_bits & ULONG_MAX,
```
llunak: The number is not supposed to be a "random" upper limit, it's supposed to be all binary ones, so the original is in fact better.

mostynb:
ULONG_MAX is all binary ones, in all C++ compiler that I'm aware of (though I don't know if it's technically safe to assume that of all standard-compliant C++ compilers).  But this is also low/no value if you're not interested in closing most/all codacy issues.

PR #359, opened 2018-01-24
[codacy] remove redundant end_status assignment

mostynb:
Variable 'end_status' is reassigned a value before the old one has been used.

### DIFF
--- a/daemon/main.cpp
+++ b/daemon/main.cpp
@@ -1391,7 +1391,6 @@ bool Daemon::handle_compile_done(Client *client)
         msg->user_msec = job_stat[JobStatistics::user_msec];
         msg->sys_msec = job_stat[JobStatistics::sys_msec];
         msg->pfaults = job_stat[JobStatistics::sys_pfaults];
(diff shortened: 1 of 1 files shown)


daemon/main.cpp
```diff
@@ -1391,7 +1391,6 @@ bool Daemon::handle_compile_done(Client *client)
         msg->user_msec = job_stat[JobStatistics::user_msec];
         msg->sys_msec = job_stat[JobStatistics::sys_msec];
         msg->pfaults = job_stat[JobStatistics::sys_pfaults];
-        end_status = job_stat[JobStatistics::exit_code];
```
HenryMiller1: Why is it redundant?  IF the job exited with a status, use that, otherwise use 151 is how I read the code.

mostynb: IIUC, line 1389 sets end_status to the value of job_stat[JobStatistics::exit_code], then we set end_status to the same value here on line 1394 (job_stat[JobStatistics::exit_code] is not modified in the meantime).

PR #361, opened 2018-01-24
[codacy] remove unneeded null-after-delete

mostynb:
Assignment of function parameter has no effect outside the function. Did you forget dereferencing it?

### DIFF
--- a/daemon/serve.cpp
+++ b/daemon/serve.cpp
@@ -313,7 +313,6 @@ int handle_connection(const string &basedir, CompileJob *job,
 
     } catch (const myexception& e) {
         delete client;
-        client = 0;
 
         if (!obj_file.empty()) {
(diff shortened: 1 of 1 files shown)


llunak:
While I do appreciate your fixes, codacy is still just a tool and there's something as taking things too far, or being simply a matter of taste. This is one of the cases where I see it as doing changes only for the sake of doing changes and satisfying a random tool and its whims. This change is in fact technically for the worse, because it removes a line that's in practice harmless (compiler it optimize it away anyway), except that it protects against a possible mistake. Being 100% codacy-perfect is just an arbitrary goal and I consider leaving this as these worth more than that.

mostynb:
Fine with me to ignore this.  Most/all of the remaining codacy issues that I didn't try to patch look useless IMO.

PR #365, opened 2018-01-25
[codacy] check if realloc failed in services/comm.cpp

mostynb:
My attempt to reset state if realloc fails needs double checking, and we should probably figure out how to return an error in the other two cases.

### DIFF
--- a/services/comm.cpp
+++ b/services/comm.cpp
@@ -90,6 +90,7 @@ bool MsgChannel::read_a_bit()
     if (count < 128) {
         inbuflen = (inbuflen + 128 + 127) & ~(size_t) 127;
         inbuf = (char *) realloc(inbuf, inbuflen);
+        assert(inbuf); // Probably unrecoverable if realloc fails anyway.
         count = inbuflen - inofs;
     }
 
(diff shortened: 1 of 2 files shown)


llunak:
This is one of the pointless Codacy checks. If the situation is so bad that realloc() for a relatively small buffer fails, the whole thing is going to tumble down anyway. No point in complicating the code just for the sake of pleasing a tool.

mostynb:
One could argue that an error in the log is valuable. But that could also fail if it needs to allocate memory on the heap.  Is it worth at least leaving source code comments?  Or asserts (I suspect not, but worth asking)?

llunak:
Adding assert() should be fine, but AFAIK that doesn't really do much in practice either. IIRC at least on Linux such small allocations simply don't fail in reality, the machine in such cases usually gets brought down to its knees by thrashing and then OOM killer just shoots something.

mostynb:
Changed to asserts.

ossilator:
assert really isn't the way to check for system errors.
also, not all systems over-commit, and even linux can be taught not to (though it's unlikely that machines participating in an icecc farm would do that).

the traditional c way to do this would be implementing a wrapper my_realloc() which aborts().

oh, and c++ has this fine concept named "exceptions" for doing non-intrusive error checking ...

mostynb:
Adding a realloc wrapper sounds reasonable to me, but I agree with @llunak that it probably wouldn't make much difference in practice.

PR #369, opened 2018-01-26
formatting fixes: ){ -> ) {

mostynb:


llunak:
Sorry, why? What's the point?

mostynb:
It makes the code slightly more readable (for me, at least).

llunak:
Well, if everybody modified the code like this to whatever feels more readable to them, git log would be full of those (for example, I dislike the idea of hiding opening brackets at the end of the line, space or not). In my experience dealing with such trivialities is just a waste of time and people are never going to be pleased about it.

PR #370, opened 2018-01-31
require C++11 to build icecc

### LINKED ISSUE #353: Is it time to bump the version number, and allow C++11 or C++14 code?
Should we bump the version number to 1.2pre1 ?

Most of the major Linux distributions seem to have at least GCC 4.9 in their supported releases, I think it should be safe for us to start allowing C++11 code, and probably even C++14 if we don't worry about Ubuntu 14.04 LTS.

Ubuntu 14.04 LTS has GCC 4.9
Ubuntu 16.04 LTS has GCC 5.3
Debian Stretch has GCC 6.3
openSUSE Leap 42.2 has GCC 6.1.1
Red Hat Enterprise Linux 6 has a "Red Hat Developer Toolset" backport of GCC 6.2 [1]
And I assume that Centos is similar to Redhat.

[1] https://access.redhat.com/solutions/19458

mostynb:
We can modernise the icecc codebase a bit now that icecc 1.1 has
been released.

GCC >= 4.8 and Clang >= 3.3 officially support C++11. Most currently
supported Linux distribution releases either have such a compiler by
default, or as an optional install:

Ubuntu 14.04 LTS has GCC 4.9
Ubuntu 16.04 LTS has GCC 5.3
Debian Stretch has GCC 6.3
openSUSE Leap 42.2 has GCC 6.1.1
Red Hat Enterprise Linux 6 has a "Red Hat Developer Toolset" backport of GCC 6.2
SUSE Linux Enterprise 11 SP4 has an optional GCC 5.2 toolchain

Resolves #353

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -1,4 +1,5 @@
 1.2 (in progress)
+ - C++11 is required to build icecc
 
 1.1
  - revert "Add load control for preprocessing"
--- a/README
+++ b/README
@@ -6,6 +6,8 @@ form of process accounting in there.
 How to install icecream
 =======================
 
+You must have a compiler that supports C++11.
+
 cd icecream
 ./autogen.sh
 ./configure --prefix=/opt/icecream
--- a/configure.ac
+++ b/configure.ac
@@ -25,6 +25,8 @@ AM_INIT_AUTOMAKE([1.11 foreign dist-bzip2])
 AM_SILENT_RULES([yes])
 AC_LANG([C++])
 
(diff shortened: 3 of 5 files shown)


HenryMiller1:
I agree with the direction, but you need to make a change that requires c++11 before we can make it mandatory.

the ::bind change is good, and we should enable C++11 if the compiler supports it, to require it is a different story.

llunak:
Current master builds with C++11, so closing this one.

PR #378, opened 2018-02-13
fix path canonicalization functions

mostynb:
get_absfilename() doesn't handle ".." correctly:
* "/foo/../bar" -> "/foo/bar" (should be "/bar")
* "/foo/..bar" -> "/foo/bar" (should be "/foo/..bar")

Let's fix this.

### DIFF
--- a/services/file_util.cpp
+++ b/services/file_util.cpp
@@ -69,50 +69,104 @@ string get_relative_path(const string &to, const string &from) {
 }
 
 /**
- * Returns a string without '..' and '.'
+ * Returns a path string with as many '..' and '.' components
+ * resolved as is possible, unambiguously.  This does not include
+ * symlink resolution.  If the input ends in '/', the output will
+ * also.
+ */
+string get_normalized_path(const string &path)
+{
+    if (path.empty()) {
+        return path;
+    }
+
+    vector<string> components = split(path, '/');
+    int up = 0; // Number of ..'s we still have to process.
+
+    // Convert resolved components to "".
+    for (int i=components.size()-1; i>=0; i--) {
+        if (components[i] == ".") {
+            components[i] = "";
+        } else if (components[i] == "..") {
+            components[i] = "";
+            up++;
+        } else if (up > 0 && components[i] != "") {
+            components[i] = "";
+            up--;
+        }
+    }
+
+    string result = "";
+
+    // If we were unable to resolve all '..' components, prepend
+    // the number that were unresolved.
+    if (path[0] != '/' && up > 0) {
+        result += "..";
+        up--;
+        for (; up>0; up--) {
+            result += "/..";
+        }
+    }
+
+    // Add back all components that weren't "resolved-away".
+    for (unsigned int i=0; i<components.size(); i++) {
+        if (components[i] != "") {
+            if (result != "") {
+                result += "/";
+            }
+            result += components[i];
+        }
+    }
+
+    // Ensure that leading and trailing slashes are retained.
+    if (path[0] == '/' && result[0] != '/') {
+        result = "/" + result;
+    }
+    if (path[path.size() - 1] == '/' && result[result.size()-1] != '/') {
+        result += "/";
+    }
+
+    return result;
+}
+
+/**
+ * Return an absolute normalized path with leading and trailing slashes
+ * preserved, unless the input is empty- in which case return the input.
+ * unchanged.
+ */
+string get_abs_path(const string &path)
+{
+   if (path.empty()) {
+        return path;
+    }
+
+    string process_me = path;
+    if (path[0] != '/') {
+        process_me = get_cwd() + '/' + process_me;
+    }
+
+    return get_normalized_path(process_me);
+}
+
+/**
+ * Returns a string without '..' and '.', and without a trailing '/'.
  *
  * Preconditions:  path must be an absolute path
  * Postconditions: if path is empty or not an absolute path, return original
  *                 path, otherwise, return path after resolving '..' and '.'
+ *                 and without a trailing '/'.
  */
 string get_canonicalized_path(const string &path) {
     if (path.empty() || path[0] != '/') {
         return path;
     }
 
-    vector<string> parts = split(path, '/');
-    vector<string> canonicalized_path;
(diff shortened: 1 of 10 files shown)


ossilator:
i'd be cautious about that - see the documentation of perl's File::Spec->canonpath().

mostynb:
The existing/broken code makes no attempt to resolve symlinks, so neither does this fix (ie this is like python's os.path.normpath rather than os.path.realpath).

We could of course consider changing this, or simply document the behaviour.

mostynb:
Oh yuck, there's also a "get_canonicalized_path" function in file_util.cpp which doesn't resolve symlinks either.  I think we should merge these two functions and resolve symlinks while we're at it.

mostynb:
Pushed a rework.  We might want to consider splitting this into two distinct functions:
1) just use realpath, when we know the path should already exist
2) resolve the largest existing subpath with realpath, and make a guess about any remaining ".." components in the remaining part of the path.

daemon/serve.cpp
```diff
@@ -240,12 +240,12 @@ int handle_connection(const string &basedir, CompileJob *job,
...
-                output_dir = tmp_path + get_canonicalized_path(file_dir);
+                relative_file_path = get_relative_path(get_abspath(job_output_file), get_abspath(job_working_dir));
+                output_dir = tmp_path + get_abspath(file_dir);
             }
             else { // output file is already relative, canonicalize in relation to working dir
-                string canonicalized_dir = get_canonicalized_path(job_working_dir + '/' + file_dir);
-                relative_file_path = get_relative_path(canonicalized_dir + '/' + file_name, get_canonicalized_path(job_working_dir));
+                string canonicalized_dir = get_abspath(job_working_dir + '/' + file_dir);
+                relative_file_path = get_relative_path(canonicalized_dir + '/' + file_name, get_abspath(job_working_dir));
                 output_dir = tmp_path + canonicalized_dir;
```
johnmiked15: There is a subtle bug here -

Changing the implementation to use `realpath` _could_ break the calculation of the working path and output file path that is used for determining where to build from and what relative path to use to the output file.

Consider the following: the host machine is attempting to build a file from one working directory, targeting the output to another directory. They send the job to icecream. The job ends up on a machine where either that working directory or the target output directory *exist as symlinks*. This code will now resolve that path to the actual path on the *build* machine, and append that to the `/tmp/` chroot folder, breaking the relative path structure needed for correct `.dwo` file paths. This actually leaks information about the build machine and breaks the resulting dwarf object information.

We truly do want to purely canonicalize the given path, *not* resolve the symlinks as well.

services/file_util.cpp
```diff
@@ -69,50 +71,49 @@ string get_relative_path(const string &to, const string &from) {
...
-    while (path_it != path_end) {
-        output += *path_it;
+    log_info() << "Path doesn't exist yet, guessing the canonical form."
+               << endl;
 
-        ++path_it;
-        if (path_it != path_end) {
-            output += "/";
+    vector<string> components = split(path, '/');
+    deque<string> stack;
```
johnmiked15: `std::vector<string>` is probably faster here, actually. In fact:

```cpp
vector<string> stack(components.size());
```
Would allocate it to be as big as it'll ever need to be so there will never be a reallocation.

daemon/serve.cpp
mostynb: > The job ends up on a machine where either that working directory or the target output directory exist as symlinks.

Is there any (non-trivial) path handling that occurs on the remote end?  I need to confirm, but aren't the remote commands handled inside a (clean) chroot?

Having said that, if we don't need to resolve symlinks then ignoring the fs is going to be faster.  I will update the PR once I find time to update and retest.

services/file_util.cpp
mostynb: We're modelling a directory stack, switching to std::vector would require the stack-ness of this to be handled manually, which would introduce some complexity.  I have a feeling that there would not be a huge performance difference in practice here- in my experience directory depths for compilation are typically under say 20.

I picked std::deque because I want to treat it like a stack during processing, but then like a queue when building the output.  It shouldn't be difficult to use std::queue instead, then build the resulting string backwards.  Does that sound efficient enough (the code should be simpler than using std::vector)?

daemon/serve.cpp
johnmiked15: This part (where it's determining the paths) is done before it chroots. It needs to figure out which directories to create inside that chroot. 

Resolving symlinks is not something this part of the code should do since it's trying to mimic the initiatior's filesystem--it shouldn't be resolving things on the remote system.

services/file_util.cpp
johnmiked15: You mean http://en.cppreference.com/w/cpp/container/vector/pop_back doesn't do what you need? I think `std::vector` is fine here.

daemon/serve.cpp
mostynb: Updated the PR, removing symlink resolution.

services/file_util.cpp
mostynb: Pushed a new version which scans through the components vector twice, without using std::deque or similar.

mostynb:
Friendly ping...

johnmiked15:
Sorry, haven't had much time recently. So before I dive back into this, can you help jog my memory? Why do we need to change the implementation of `get_canonicalized_path()`? It appears like the new `get_normalized_path()` does the same thing, but the implementation is quite different. I'm in favor of keeping the implementation as it was before.

Speaking of introducing another path normalization function, have we considered just fixing the usages everywhere so they only need one form or the other? It is very confusing at first glance to have `get_canonicalized_path()` *and* `get_normalized_path()` that each do something slightly differently. These should either be merged (and usage code fixed), or they need to be better named.

llunak:
I've merged request #403 , which fixes the /../ problem. If there are still other problems pending, please reopen and describe them and open a new issue for them.

bwiklak:
There are still problems with path canonicalization.
with `-fsanitize-blacklist=../../d/blacklist.txt` it should resolve to `/a/d/balcklist.txt` but is resolved to `/a/b/c/d/blacklist.txt` instead.

PR #380, opened 2018-02-19
Do not send clang plugins with each compile job

pr0t4zy:
This will not work anyway in case of cross-compiling. Also, there is no need to actually have the plugin in the path, having it reachable by dlopen() is enough (but should we check it?).

### DIFF
--- a/client/arg.cpp
+++ b/client/arg.cpp
@@ -608,20 +608,8 @@ bool analyse_argv(const char * const *argv, CompileJob &job, bool icerun, list<s
                         if (argv[i + 1] && argv[i + 2] && str_equal(argv[i + 1], "-Xclang")) {
                             args.append(a, Arg_Rest);
                             args.append(p, Arg_Rest);
-                            string file = argv[i + 2];
-
-                            if (access(file.c_str(), R_OK) == 0) {
-                                file = get_absfilename(file);
-                                extrafiles->push_back(file);
-                            } else {
-                                always_local = true;
-                                log_info() << "plugin for argument "
-                                           << a << " " << p << " " << argv[i + 1] << " " << file
-                                           << " missing, building locally" << endl;
-                            }
-
                             args.append(argv[i + 1], Arg_Rest);
-                            args.append(file, Arg_Rest);
+                            args.append(argv[i + 2], Arg_Rest);
                             i += 2;
                         }
                     } else {


llunak:
It will work in the case of not cross-compiling, which is good enough, and this change breaks it, as can be seen by the failing test. What is the problem you're trying to solve? Whatever it is, this is not the fix.

pr0t4zy:
I am trying to cross-compile with clang plugins enabled. The tests seem to be testing the non cross-compiling scenario, maybe it would be better not to break it indeed.

What I want to achieve:
1. Put clang plugins in env package, under e.g. /usr/lib (so that it can be dlopen'ed)
2. Pass filename (with no path) to -Xclang -load, this way remote hosts will be able to use the plugin
3. For local compilation we can either use also filename-only and modify LD_LIBRARY_PATH (this would be up to the caller) or keep whatever we get (usually /some/path/to/plugin.so) and strip the path for remote jobs.

Then we could use something like ICECC_CLANG_PLUGINS_IN_ENV to switch between two behaviors.
Does it sound ok to you?

pr0t4zy:
Moreover, I think it makes much more sense to send plugins in env package, as they will not change for each compile job. It would save a *lot* of time.

llunak:
Plugins already do get sent as part of the env tarball, they are included by using icecc-create-env's --addfile argument. I do not see anything wrong with the non-crosscompiling case.

If you want cross-compiling to work with plugins, it should work to add the plugin manually using --addfile to the env tarball that you have to manually create. And in case that doesn't work for you for some reason, that's all that should need fixing.

pr0t4zy:
They do get sent if you use --addfile, I am not denying that (and actually my patch relies on that behavior, see 1. in my previous comment).

The problem here is: even if you do prepare you env tarball with plugins inside, the plugins will be sent anyway with each compile job (also for non-crosscompiling case!), and this is a waste. Also, you will need to put the plugins in the same directory structure in the env tarball as you have them locally, which can be problematic in case the plugins are kept inside source tree (like it is done with chromium).

I propose to rewrite the plugin path for remote jobs from /path/to/my/plugin.so into plugin.so, and put plugin.so in usr/lib in env tarball. Like this clang will still be able to find the plugin for local job (as we will leave /path/to/my/plugin.so) and for remote jobs plugin.so (with no path) will be used, and will work because of plugin.so being in usr/lib.

mostynb:
@pr0t4zy, is this the use-case you're talking about: user1 and user2 both build toolchain tarballs with the same toolchain and same clang plugins, but with different paths to the plugins (eg inside /home/user1/ somewhere and /home/user2/ somewhere), then these two toolchain tarballs have different hashes and are distributed separately.

BTW, I think there are also path handling issues with chromium's clang plugins, since the paths are relative and that is currently broken (I'm trying to work through that in PR #378).

mostynb:
I debugged this a bit, and tried to put together a descriptive issue: #384

pr0t4zy:
Sorry for the delay. Yes @mostynb, this is the case I am talking about. Thanks for opening the issue, it summarizes everything nicely. I guess we can move the discussion there.

liljenzin:
#388 provides a more flexible solution to the same problems as an environment variable can be set to rewrite individual plugin paths, thus not changing the default behavior.

pr0t4zy:
Yes that is much better. Closing this PR.

PR #381, opened 2018-02-27
Revert "Fix building with precompiled headers."

rgriebl:
This "fix" is wrong since it does exactly the opposite of what the commit
message says. The code was correct in the first place - access() will return
a false value (0) on success and a true value (-1) on error.
So if(access) has to be read as if(not_exists).

This reverts commit 5b27d706ea430ff1a42a6aec489b2790056a07d4.

### DIFF
--- a/client/arg.cpp
+++ b/client/arg.cpp
@@ -458,9 +458,8 @@ bool analyse_argv(const char * const *argv, CompileJob &job, bool icerun, list<s
                     if (dot_index != string::npos) {
                         string ext = p.substr(dot_index + 1);
 
-                        if (ext[0] != 'h' && ext[0] != 'H'
-                            && (access(p.c_str(), R_OK)
-                                || access((p + ".gch").c_str(), R_OK))) {
+                        if (ext[0] != 'h' && ext[0] != 'H' && access(p.c_str(), R_OK)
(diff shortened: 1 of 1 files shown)


llunak:
You are right that 5b27d70 is wrong, but simply reverting it will still keep around the original problem it was trying to fix. Can you instead submit a patch that fixes the access() usage?

rgriebl:
There was no problem as far as I can see. The check is fine as it was: not_exists(a) && not_exists(b) only bails out if both a and b are missing. I have no idea why this "fix" was committed in the first place. We just did a couple of Qt re-builds with my revert and we're back to nice -j 100 performance again.

ossilator:
i do see a problem, which is the terrible style. use access(...) < 0 for the error check, and the code becomes a lot clearer.

rgriebl:
Agreed - doing POSIX result checks like this is bad style, but it would still be lot cleaner (especially history wise) to first revert the bogus commit and THEN follow up with a cleanup (if desired).

ossilator:
no, it wouldn't. reverts are what noises up the history. always go right for the end state, fixing style as a side effect on the touched lines.

rgriebl:
That's your personal opinion that I do not share -- looking at icecream's git log I'd say Luboš is the one to decide for the icecream project.

Issue #379, opened 2018-02-14
Should we add some unit tests?


mostynb:
I think it would be good if the test suite included some unit tests.  Would this be welcome?  My preference would be to use gtest.

llunak:
There actually already is one unit test, tests/args.cpp, it's just practically hidden by the scripted tests, those should get separated into different directories. But otherwise, if you think there's something to test, sure, feel free to add tests.

llunak:
The question has been answered, and this is way too generic to do something more about this. If there's a specific test to be added, it just should be added, but not much point in this generic issue.

PR #383, opened 2018-03-12
"make test" should build and run unittests too

mostynb:
We have a small number of unit tests, but it seems that they are not
run by default and have become a little stale.

Some background in #379.

This is a WiP- my automake knowledge is pretty nonexistent, I'm sure that this part will need cleanup.

### DIFF
--- a/client/local.cpp
+++ b/client/local.cpp
@@ -199,7 +199,7 @@ bool compiler_only_rewrite_includes(const CompileJob &job)
 
 string clang_get_default_target(const CompileJob &job)
 {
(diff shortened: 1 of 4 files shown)


spacemoose:
My automake knowledge is essentially zero.  Would there be a willingness to use cmake for the build?  I would be interested in doing the work.

HenryMiller1:
@spacemoose my auto tools knowledge is minimal as well. I started working on cmake, but lost interest. My code is in my fork, and someone has done a pull request there, but it isn't really functional. Feel free to peak if you like.

llunak:
I have handled this while doing other work, thank you.

PR #388, opened 2018-03-28
Plugin alias

liljenzin:
The patches add support for rewriting remote paths for clang plugins using a new environment variable ICECC_PLUGIN_ALIAS. Rewriting the plugin paths are needed when clang plugins are loaded from relative paths and the same tarball should be used to build different worktrees.

### DIFF
--- a/README.md
+++ b/README.md
@@ -414,7 +414,7 @@ the tarball path is used instead of the local path to load the plugin:
 
       export ICECC_PLUGIN_ALIAS=../plugins/libFoo.so=usr/lib/libFoo.so
 
-Plugin aliases are only used when compiling with clang and has no effect
+Plugin aliases are only used when compiling with clang and have no effect
 when gcc is used as compiler.
 
 Debug output
--- a/README.md
+++ b/README.md
@@ -36,6 +36,8 @@ Table of Contents
     icecream](#cross-compiling-for-multiple-targets-in-the-same-environment-using-icecream)
 -   [How to combine icecream with
     ccache](#how-to-combine-icecream-with-ccache)
+-   [Using plugin aliases to rewrite remote plugin
+    paths](#using-plugin-aliases-to-rewrite-remote-plugin-paths)
 -   [Debug output](#debug-output)
 -   [Some Numbers](#some-numbers)
 -   [What is the best environment for
@@ -384,6 +386,36 @@ And then compile with
 Note however that ccache isn't really worth the trouble if you're not
 recompiling your project three times a day from scratch (it adds some
 overhead in comparing the source files and uses quite some disk space).
+
+Using plugin aliases to rewrite remote plugin paths
+-------------------------------------------------------------------------
+
+Plugin aliases can be used to rewrite remote plugin paths so that the
+remote compiler can load them from the unpacked tarball using different
+paths:
+
+      ICECC_PLUGIN_ALIAS=<local_name1>=<remote_name1>,<local_name2>=...
+
+When the variable ICECC_PLUGIN_ALIAS is set to a comma-separated list of
+aliases, the aliases will be used to replace local plugin paths by remote
+plugin paths.
+
+In the following example the compiler loads the plugin libFoo.so from a
+path relative to the current worktree:
+
+      clang -Xclang -load -Xclang ../plugins/libFoo.so -c bar.so
+
+Assume the plugin was added to the tarball with the following path:
+
+      usr/lib/libFoo.so
+
+The following alias definition will rewrite remote parameters so that
+the tarball path is used instead of the local path to load the plugin:
+
+      export ICECC_PLUGIN_ALIAS=../plugins/libFoo.so=usr/lib/libFoo.so
+
+Plugin aliases are only used when compiling with clang and has no effect
+when gcc is used as compiler.
 
 Debug output
 -------------------------------------------------------------------------
--- a/client/arg.cpp
+++ b/client/arg.cpp
@@ -33,6 +33,8 @@
 #include <sys/types.h>
 #include <sys/stat.h>
 
+#include <map>
+
 #include "client.h"
 
 using namespace std;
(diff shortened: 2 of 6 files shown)


README.md
```diff
@@ -385,6 +387,36 @@ Note however that ccache isn't really worth the trouble if you're not
...
+Assume the plugin was added to the tarball with the following path:
+
+      usr/lib/libFoo.so
+
+The following alias definition will rewrite remote parameters so that
+the tarball path is used instead of the local path to load the plugin:
+
+      export ICECC_PLUGIN_ALIAS=../plugins/libFoo.so=usr/lib/libFoo.so
+
+Plugin aliases are only used when compiling with clang and has no effect
```
mostynb: nitpick: s/has/have/

mostynb:
This seems like a reasonable solution to me.  Perhaps the path handling change can be replaced by my pull request #378?

liljenzin:
@mostynb I didn't know about you pull request and as soon as #378 gets merged I will drop my own patch for path normalization and rebase this pull request on the new master.

Of course, the algorithm I used is both faster and cleaner as it scans the string once and produces the output on the fly, instead of creating a stack of components that you iterate over several times and all of that. On the other hand I don't think argument parsing will have any visible impact at all on overall performance and I don't really care about how it is done as long as the output is correct.

honwsn:
like "export ICECC_PLUGIN_ALIAS=../plugins/libFoo.so=usr/lib/libFoo.so"

 "usr/lib/libFoo.so" just in tarball that added by addfile option.

, how "usr/lib/libFoo.so" been found in remote compile node?

libFoo.so in remote compile node may be like this 
"./var/cache/icecc/target=x86_64/c391475aa1b0b4c16595945c356fea6b/usr/lib/libFoo.so"

liljenzin:
@honwsn The remote node will run the compiler in a change-rooted environment.
https://en.wikipedia.org/wiki/Chroot

honwsn:
thanks @liljenzin   I use your patch on newest  code , compile and run icecc,  the path replace for remote node such as usr/lib/libFoo.so **did  effect.** 

you'd better think about relative path in the key.  rebase path and remove  ".." ,
use absoute path as key.

I find a problem in icecream
when set clang_use_chrome_plugins as true ( it's a default value),   will get following logs 

"ICECC[6489] 17:33:36: <write_server_cpp from cpp>
ICECC[6489] 17:33:36: sent 143201 bytes (28%)
ICECC[6489] 17:33:36: </write_server_cpp from cpp: 118ms>
ICECC[6489] 17:33:36: <wait for cpp>
ICECC[6489] 17:33:36: </wait for cpp: 0ms>
ICECC[6489] 17:33:36: <wait for cs>
ICECC[6489] 17:34:36: saw eof without complete msg! 1
ICECC[6489] 17:34:36: </wait for cs: 60261ms>
ICECC[6489] 17:34:36: saw eof without complete msg! 1
ICECC[6489] 17:34:36: got exception Error 14 - error reading message from remote"



so i think clang_use_chrome_plugins did not support by icecream  as so far and your patch cannot help me to fix chrome plugins problem , for example use libFindBadConstructs.so plugin to regular team's code to follow  chrome style.

liljenzin:
@honwsn Try the following to build chromium:
```
icecc-create-env $CHROMIUM_DIR/third_party/llvm-build/Release+Asserts/bin/clang --addfile $CHROMIUM_DIR/third_party/llvm-build/Release+Asserts/lib/libFindBadConstructs.so --addfile $CHROMIUM_DIR/third_party/llvm-build/Release+Asserts/lib/libBlinkGCPlugin.so
export ICECC_PLUGIN_ALIAS=../../third_party/llvm-build/Release+Asserts/lib/libFindBadConstructs.so=/usr/lib/libFindBadConstructs.so,../../third_party/llvm-build/Release+Asserts/lib/libBlinkGCPlugin.so=/usr/lib/libBlinkGCPlugin.so
```

honwsn:
@liljenzin Thank for your reply. 

I had did  it before like your post .  libBlinkGCPlugin  not be used as so far.   the clang path replacement  is right for remote compile task，such as libFindBadConstructs use  /usr/lib/libFindBadConstructs.so 
but  got exception "Error 14 - error reading message from remote " in remote node, or remote node had no reponse for a long time  when I set clang_use_chrome_plugins as true .  I  test these  in cross compile enviroment for chrome android.

honwsn:
OK,  I add libBlinGCPlugin  to prevent anything else.  
when clang compile  remote failed , it just run locally
"sh: 1: -dumpmachine: not found
ICECC[25656] 20:47:34: failed to read default clang host platform, building locally
"

liljenzin:
@honwsn I'm not convinced the first problem you mention is related to to the plugin patch. There are other reasons for why some of the v8 compilations end up running locally, also when not using plugins.

The second problem is because of a bug in the plugin itself. libFindBadConstructs uses a list of hard coded path patterns to exclude generated files from style checking, but the patterns will no longer match exactly after the source files have been rebased to the change rooted file tree on the remote host. You can work around this problem be adding the following flags to the compiler:
`-Xclang -plugin-arg-find-bad-constructs -Xclang no-realpath`

honwsn:
@liljenzin  Thanks for your suggestion,  whether chromium style check will go on for normal fiels  if set ``-Xclang no-realpath` ?

liljenzin:
@honwsn The style checking should work as normal also with this flag. It just relaxes the pattern matching so that it works with rebased file names.

honwsn:
@liljenzin  I test it as you said ,  That  did fix the chromium style check in icecc ,  perfect solution !  **Thanks for your patch and suggestion.**

llunak:
The patch seems technically ok, but I don't understand the use case. Can you provide a specific example of when one would want to use this feature?

liljenzin:
Closing as chromium changed to link the plugin libraries statically into clang, removing the need for special handling.

PR #392, opened 2018-04-22
Support adding gcc and clang to the same environment

pprkut:
The necessary support code is already present, it was just missing the entry points that would allow having both.

I tested the resulting environment package and confirmed that compiling succeeded with both clang (6.0.0) and gcc (7.3.0).

### DIFF
--- a/client/icecc-create-env.in
+++ b/client/icecc-create-env.in
@@ -193,11 +193,25 @@ if test "$1" = "--gcc"; then
     added_gxx=$1
     shift
     gcc=1
+    if test "$1" = "--clang"; then
+        shift
+        added_clang=$1
+        shift
+        if test "x$1" != "x--addfile" -a "x$1" != "x--gcc" -a -e "$1"; then
+            # accept 2nd argument being the compilerwrapper binary, for backwards compatibility
+            added_compilerwrapper=$1
+            shift
+        fi
+        if test -z "$added_compilerwrapper"; then
+            added_compilerwrapper=@PKGLIBEXECDIR@/compilerwrapper
+        fi
+        clang=1
+    fi
 elif test "$1" = "--clang"; then
     shift
     added_clang=$1
     shift
-    if test "x$1" != "x--addfile" -a -e "$1"; then
+    if test "x$1" != "x--addfile" -a "x$1" != "x--gcc" -a -e "$1"; then
         # accept 2nd argument being the compilerwrapper binary, for backwards compatibility
         added_compilerwrapper=$1
         shift
@@ -206,6 +220,14 @@ elif test "$1" = "--clang"; then
         added_compilerwrapper=@PKGLIBEXECDIR@/compilerwrapper
     fi
     clang=1
+    if test "$1" = "--gcc"; then
+        shift
+        added_gcc=$1
+        shift
+        added_gxx=$1
+        shift
+        gcc=1
+    fi
 else
     if test -z "$1"; then
         usage


mostynb:
What is the use-case for this?

pprkut:
I'm a distro-packager and compile lots of software on my local infrastructure. Not all sources use the same compiler though, some use gcc and some prefer or even require clang. That's the one side of it.
The other is that ICECC_VERSION doesn't support specifying more than one compiler environment per architecture. I'd need to constantly alter that variable's content before starting the compile to point to either clang or gcc packages, which is quite a hassle and hinders automation.
My current approach is a symlink that I switch between gcc and clang environments and use in ICECC_VERSION. That, however, means that everytime I switch I have to manually purge the environment from all nodes because the name is the same.

All in all it's not a huge issue to not have this, but it's removing quite some obstacles to make it a smoother experience, and it's optional still after all.

mostynb:
Another trick that you might consider is to use shell scripts to automate the switch.

For example, "source setup_icecc_clang.sh" could set ICECC_VERSION to point to a clang tarball, and you could have a different script for each toolchain you want to use.  You could possibly also implement this as a wrapper script which just sets environment variables before exec'ing a specified command, like "icecc_clang.sh make ..."

pprkut:
I already have a shell script to set up the environment appropriately, that's where part of the difficulty stems from. I'd really need different tooling for clang and gcc in an area that shouldn't have to know anything about the compiler being used.
And in scenarios where I switch between compiling with clang and gcc for different builds it would still mean manual interaction to switch also the tooling, which is tricky considering I may not know upfront which compiler is being used.
From my perspective having gcc and clang in the same environment package would make this all so much simpler.

HenryMiller1:
I would have expected this to work already...

That said, is this the right approach?

As I look at my environment, I'm not really interested in managing ICECC_VERSION when I have gcc, gcc-5.2, gcc-6, gcc-7, clang-3.9, clang-5.0.1, clang 6.0... all installed, plus cross compilers for a few others embedded systems I build for.  The nightmare gets worse for those who don't have a monolithic CPU architecture they build on.  I think a smaller configuration file that icecream reads of some sort would be better.

I'm also not sure I like this - I really want ice cream to be smart enough to know if it  needs to send the gcc compiler or the clang compiler across the wire. Disk space is cheap these days, but it isn't free, and networks are still slow, so not sending gcc to a host that needs clang could potentially speed up the build. (I wonder if just sending g++ and not gcc would help - looks like on my machine this could save half a meg or so...

pprkut:
Oh sure, I took the easy way out. This gives me an instant improvement for my workflow :)
ICECC_VERSION is nice, because I can use a pre-generated environment to send across the wire and don't have to wait for it to be generated on the fly. The size isn't really an issue for me since it's shipped once and then remains on that machine until it reboots or I manually purge it, either way it's probably a couple weeks.
Now, a configuration file could certainly work as well. But maybe simply standardizing a naming scheme for the environment packages would be enough? Then you'd only need to specify the location where to look for pre-generated environments and icecream could look for the existance of those on demand and send a matching one over the wire, or create one on demand if it doesn't exist, just like it's doing now. The naming scheme could look like this:

${OS}-${TARGET_ARCH}-${HOST_ARCH}-${COMPILER}-${COMPILER_VERSION}

For my setup this would get me:

slackware-x86_64-x86_64-gcc-7.3.0
slackware-x86_64-x86_64-clang-6.0.0

HenryMiller1:
Opened #395 to track something better. This works for now.

Issue #300, opened 2017-10-10
Build conftest.c locally


tbe:
To avoid blocking of hosts and long stale times during the execution of autoconf scripts, `conftest.c` should be forced to compile on the local system.

See https://github.com/distcc/distcc/blob/848e4be57bd1c3db30bd1b767660736d77585ca9/src/filename.c#L293

llunak:
While this sort of makes sense, in practice I don't see the point of this. Do you have such problems often? If yes, then you should have exactly the same kind of problems during normal compiles. And in that case what's the point of special-casing just one instance of the problem?

HenryMiller1:
It makes sense if it is more generic. The overhead of ice cream seems to be 100ms. Autoconf is single threaded so every 10 try compiles will add a second of ice cream overhead. These are all trivial files that build in a few ms, so the difference is noticeable overall. 

Cmake has the same problem. Many try complies of trivial files. I would assume that all other portable buildsystems do as well.  Thus I don't want to special case Autoconf, I want a generic way to know if we are in a single threaded configuration phase. This means should be shared with distcc, so any proposal needs to be discussed with them. 

For now I export ICECE=no when running cmake, it makes a big difference.

tbe:
Also, as many autoconf tests may fail, it seems like icecc is marking the executing node as down.

coolo:
doing so for performance overhead makes sense, but icecc shouldn't care if the complicated failed or not. If that's the case (for you) that's a different issue.

llunak:
https://github.com/icecc/icecream/pull/402 has implemented this for configure. I haven't found a good way to detect icecc is called from cmake, so unless somebody knows, there's not much more to do. As this is only a performance improvement, closing.

PR #402, opened 2018-07-12
Always compile autoconf tests locally

kito-cheng:
Compile locally if it seems like a autconf test.

#300

### DIFF
--- a/client/arg.cpp
+++ b/client/arg.cpp
@@ -55,6 +55,22 @@ inline int str_startswith(const char *head, const char *worm)
     return !strncmp(head, worm, strlen(head));
 }
 
+/* Some files should always be built locally... */
+static bool
+should_always_build_locally(const string &filename)
+{
+    string p;
+
+    p = find_basename(filename);
+
+    if (str_startswith("conftest.", p.c_str())
+        || str_startswith("tmp.conftest.", p.c_str())) {
+        return true;
+    }
+
+    return false;
+}
+
 static bool analyze_program(const char *name, CompileJob &job, bool& icerun)
 {
     string compiler_name = find_basename(name);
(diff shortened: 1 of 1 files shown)


client/arg.cpp
```diff
@@ -55,6 +55,44 @@ inline int str_startswith(const char *head, const char *worm)
     return !strncmp(head, worm, strlen(head));
 }
 
+/**
+ * Return a pointer to the basename of the file (everything after the
+ * last slash.)  If there is no slash, return the whole filename,
+ * which is presumably in the current directory.
+ **/
+static const char *find_basename(const char *sfile)
```
llunak: There already is find_basename() function (utils.h).

client/arg.cpp
```diff
@@ -55,6 +55,44 @@ inline int str_startswith(const char *head, const char *worm)
...
+
+    if (slash == NULL || slash[1] == '\0')
+        return sfile;
+
+    return slash + 1;
+}
+
+/* Some files should always be built locally... */
+static bool
+source_needs_local(const string &filename)
```
llunak: Source needs local what? Please use a more descriptive name, e.g. should_always_build_locally() or somesuch.

client/arg.cpp
```diff
@@ -55,6 +55,44 @@ inline int str_startswith(const char *head, const char *worm)
...
+{
+    const char *p;
+
+    p = find_basename(filename.c_str());
+
+    if (str_startswith("conftest.", p) || str_startswith("tmp.conftest.", p)) {
+#if CLIENT_DEBUG
```
llunak: It is fine to always have this info message, the #if shouldn't be there.

client/arg.cpp
```diff
@@ -55,6 +55,44 @@ inline int str_startswith(const char *head, const char *worm)
...
+        log_info() << "autoconf tests are run locally: " << p << endl;
+#endif
```
llunak: The message should be shown normally (if debug is set to the info level), the #if shouldn't be there.
Also, the right place for this message should be where the local build is forced (i.e. the "always_local = true" in the other change), as that is the place where the action takes place. If this function got called from multiple place for whatever reason, it'd print the message multiple times.

llunak:
Generally looks good, but could you please first do the suggested cleanups?

kito-cheng:
Thanks your review, I've updated according your comment :)

llunak:
Ok, thank you for the patch.

PR #403, opened 2018-07-17
Fix get_absfilename()

hferreiro:
Previously, `get_absfilename()` was replacing '/..' for '/', without
deleting the parent directory string.


liljenzin:
Also #388 contains an isolated patch that fixes the same issue, but in a more efficient way as it doesn't have to scan the input string multiple times.

PR #406, opened 2018-07-19
Always compile cmake tests locally

kito-cheng:
It's hard code to check the file name which might be used in cmake[1], the list is incomplete
 but I think it should cover most case.

[1] https://gitlab.kitware.com/cmake/cmake/tree/master/Modules  (*.c.in)

### DIFF
--- a/client/arg.cpp
+++ b/client/arg.cpp
@@ -57,15 +57,34 @@ inline int str_startswith(const char *head, const char *worm)
 
 /* Some files should always be built locally... */
 static bool
-should_always_build_locally(const string &filename)
+should_always_build_locally(const string &filepath)
 {
     string p;
 
-    p = find_basename(filename);
+    p = find_basename(filepath);
+    const char *filename = p.c_str();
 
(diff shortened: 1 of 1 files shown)


client/arg.cpp
```diff
@@ -57,17 +57,36 @@ inline int str_startswith(const char *head, const char *worm)
...
+        "CheckIncludeFile.",
+        "CheckFunctionExists.",
+        "CheckSymbolExists.",
+        "CheckTypeSize.",
+        "CheckPrototypeDefinition.",
+    };
+
+    /* cmake */
+    if (str_startswith("Check", filename)) {
+        for( size_t i = 0; i < sizeof( cmake_checks ) / sizeof( cmake_checks[ 0 ] ); ++i ) {
```
HenryMiller1: It looks to me like you just wrote std::find_if.  I'd recommend using the standard algorithm instead.

llunak:
It looks like we're not going to get the asked for tweak, and it's minor detail, so let's merge this in as it is. Thank you for the patch.

PR #407, opened 2018-07-23
in flush_writebuf send no more than 10k bytes at once, helps on slow networks

mvslobbe:
I tend to run icecc to kick off hundreds of icecc jobs in parallel. When I do this and all jobs flood the network at once, more often than not some of these jobs will get stuck - they linger for the 20 seconds timeout that has been defined and will continue to be built locally after.

I found that when I just reduce the number of bytes we write, jobs don't get stuck nearly as often any more. The flip-side of this is of course that misbehaving hosts can make a job takes ages - if one remote hosts consumes incoming data at 1byte/19 seconds we will happily wait for a very long time for the transfer to succeed.

Still - this change had made my situation a whole lot better and I suggest it makes it into the main branch.

### DIFF
--- a/services/comm.cpp
+++ b/services/comm.cpp
@@ -69,6 +69,14 @@ using namespace std;
 
 #define MAX_MSG_SIZE 1 * 1024 * 1024
 
+/*
+ * On a slow and congested network it's possible for a send call to get starved.
+ * This will happen especially when trying to send a huge number of bytes over at
+ * once. We can avoid this situation to a large extend by sending smaller
+ * chunks of data over.
+ */
+#define MAX_WRITE_SIZE 10 * 1024
+
 /* TODO
  * buffered in/output per MsgChannel
     + move read* into MsgChannel, create buffer-fill function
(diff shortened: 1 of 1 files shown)


llunak:
I assume that the real problem has been fixed with #486 , so I intend to revert this change as by now that should be just a needless slowdown. Can you please verify that the change (or current git icecream) help you as well.

mvslobbe:
You're right - #486 does address my biggest concern - that after flooding the network and timing out many jobs would return to the local machine and would cause a flood there as well. And I appreciate how #407 is a compromise, where you would sacrifice some throughput in order to better handle slow networks. But I think that in our situation this is still a compromise worth taking - I still believe that with a large number of nodes ( and potentially network slowness ) you would achieve a higher throughput if you allow for some slowness on the network, rather than resorting to giving up and compiling locally instead.

llunak:
Fair enough. I'll just put the small send size behind some ICECC_SLOW_NETWORK=1 that'll be disabled by default.

PR #412, opened 2018-08-07
Load calculation and job distribution improvements

lynxeye-dev:
This pull-request does the following:
- Simplify the CPU load calculation a bit.
- Changes the memory load calculation to have an effect again (currently it would just report load 0 if there is more than 128MB of free RAM, which is very common with modern machines)
- Changes the job distribution to spread jobs a bit more if machines are roughly of the same speed, instead of fully loading the fastest machine (this works better on modern CPUs with SMT and TurboMode where the system is faster if it isn't fully loaded, but throughput still increases if enough jobs are available for distribution).
- Distribute a small number of probe jobs to clients that haven't seen a job in a while. This allows compile nodes to recover their speed rating, if they got downrated due to external influences outside of the control of icecc. Currently if the cluster is never fully loaded, build nodes with a bad speed rating never get any jobs assigned and are thus unable to get a new (possibly better) speed rating calculated.

We are running those changes internally for a while already. Exact performance numbers are hard to come up with due to a lot of noise on the cluster, but our results seem to indicate a reduction in overall build times of about 30% (use-case is mostly build of complete system rootfs via embedded build systems like ptxdist).

### DIFF
--- a/daemon/load.cpp
+++ b/daemon/load.cpp
@@ -204,11 +204,7 @@ static void updateCPULoad(CPULoadInfo *load)
         load->userLoad = (1000 * (currUserTicks - load->userTicks)) / totalTicks;
         load->sysLoad = (1000 * (currSysTicks - load->sysTicks)) / totalTicks;
         load->niceLoad = (1000 * (currNiceTicks - load->niceTicks)) / totalTicks;
-        load->idleLoad = (1000 - (load->userLoad + load->sysLoad + load->niceLoad));
-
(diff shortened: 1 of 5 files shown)


HenryMiller1:
I think the codacy error is worth fixing.  Otherwise this looks better than what we had before.

lynxeye-dev:
I've updated the pull-request to fix the real codacy issues. The remaining one is a false-positive I believe, it seems that codacy doesn't know how to handle sysctlbyname() or scan_one().

Issue #424, opened 2018-10-23
Reordering of compiler arguments causes wrong debug information level


EnricoB1:
icecream is reordering some arguments, which causes issues when trying to build with reduced debug information. 

Tested with gcc 5.2. Building with -gsplit-dwarf and reduced debug information -g1:

`gcc -c foo.cpp -o foo.o -gsplit-dwarf -g1`

but icecream will change this to something like:

`gcc -c foo.cpp -o foo.o -g1 -gsplit-dwarf`

The result is a binary with full debug info.

HenryMiller1:
I think this is fixed, if not re-open

EnricoB1:
I ran the tests added in pull request #425 and the issue is not fixed. See [test run](https://travis-ci.org/EnricoB1/icecream/builds/461163347). Looks like it is fixed for the local compile, but not for remote compile. Since I can't re-open an issue closed by a collaborator, I'll have to create a new one.

PR #425, opened 2018-10-23
Keep order of compiler arguments (#424)

EnricoB1:
Fixing issue #424 . Especially to fix reordering of -g<level> and -gsplit-dwarf compiler arguments.

### DIFF
--- a/daemon/serve.cpp
+++ b/daemon/serve.cpp
@@ -317,6 +317,8 @@ int handle_connection(const string &basedir, CompileJob *job,
         if (rmsg.status == 0) {
             write_output_file(obj_file, client);
             if (rmsg.have_dwo_file) {
+                // optional flag required because depending on its arguments, the
+                // compiler might not write a dwo file.
                 write_output_file(dwo_file, client, true /* optional */);
             }
         }
--- a/daemon/serve.cpp
+++ b/daemon/serve.cpp
@@ -81,14 +81,21 @@ error_client(MsgChannel *client, string error)
     }
 }
 
-static void write_output_file( const string& file, MsgChannel* client )
+static void write_output_file( const string& file, MsgChannel* client, bool optional = false)
 {
     int obj_fd = -1;
     try {
         obj_fd = open(file.c_str(), O_RDONLY | O_LARGEFILE);
 
         if (obj_fd == -1) {
-            log_error() << "open failed" << endl;
+            if(optional) {
+                if(!client->send_msg(EndMsg())) {
+                    log_info() << "write of obj end failed " << endl;
+                    throw myexception(EXIT_DISTCC_FAILED);
+                }
+                return;
+            }
+            log_error() << "open failed for: " << file << endl;
             error_client(client, "open of object file failed");
             throw myexception(EXIT_DISTCC_FAILED);
         }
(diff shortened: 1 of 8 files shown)


HenryMiller1:
I'd like to see a test. I know writing tests for icecream is hard, but without a test someone is likely to break this again in the future.

services/job.cpp
```diff
@@ -28,13 +28,14 @@
 
 using namespace std;
 
-list<string> CompileJob::flags(Argument_Type argumentType) const
```
HenryMiller1: I hate to lose type safety.

EnricoB1:
Added a test and also checked that it is indeed failing without the fix.

services/job.cpp
EnricoB1: Yes, I felt so too and though about it for some time. But I could not find a good solution. 
I could revert the ArgumentType enum to not use flags and implement customized for loops for remoteAndRestFlags() and restFlags(). That would cause some code duplication, but keep the type safety.
Should I do that instead, or do you have an alternative solution?

EnricoB1:
While trying to reproduce the test failures, I found an additional issue. clang does not write a dwo file when reducing debug output. So `clang test.cpp -gsplit-dwarf -g1` doesn't output a dwo file. Since icecream always assumed a dwo file is present if "-gsplit-dwarf" is given, i had to fix that.

services/job.h
```diff
@@ -29,9 +29,10 @@
 #include <sstream>
 
 typedef enum {
-    Arg_Local,  // Local-only args.
-    Arg_Remote, // Remote-only args.
-    Arg_Rest    // Args to use both locally and remotely.
+    Arg_Local = 0x1,               // Local-only args.
+    Arg_Remote = 0x2,              // Remote-only args.
+    Arg_RestAndDwarfFission = 0x4, // like Arg_Rest, but only if DwarfFission is enabled.
```
llunak: Why is there a separate type for -gsplit-dwarf? As far as I can judge it can be normally included in Arg_Rest.

daemon/serve.cpp
```diff
@@ -310,7 +317,7 @@ int handle_connection(const string &basedir, CompileJob *job,
         if (rmsg.status == 0) {
             write_output_file(obj_file, client);
             if (rmsg.have_dwo_file) {
-                write_output_file(dwo_file, client);
+                write_output_file(dwo_file, client, true /* optional */);
```
llunak: Please add a comment for why it's optional here, so that one doesn't need to search git history when looking at the code later.

EnricoB1: +1

services/job.h
EnricoB1: It is required to keep the current behavior of adding -gsplit-dwarf only when compiling. The flag is dropped in all other cases. It looks like this was added intentionally, so I tried to keep it. I guess this is for clang, which complains when adding the flag without compiling.

llunak: I see. But if the flag is to be used only for compiling, then it should work to keep it in Arg_Remote (which is a bit of a misnomer, Arg_Local technically means "for preprocessing" and Arg_Remote means "for compiling"). Does it work that way?

EnricoB1:
@llunak Should I put the commit mentioned in the services/job.h discussion into this pull request?

EnricoB1:
Created new issue #435 for the remaining reordering of remote compile arguments. See comment in #424

llunak:
I apologize about the late reply. I've created request #477 based on your pull request, which should fix these problems.

PR #432, opened 2018-11-20
Add support for Zstandard compression of the content

thiagomacieira:
Zstandard (https://zstd.net) is a modern compression library also of the
Lempel-Ziv family of compressors. Compared to LZO, if given the same CPU
time, zstd will compress better. Testing with the "lzop" and "zstd"
command-line tools indicates that the same-CPU-time level compression
for a 7241 kB preprocessed source lies between zstd levels 2 and 3 (LZO
takes 916 µs; zstd level 2 takes 753 µs and level 3 takes 1024 µs).
Level 2 compresses this file to 1028 kB, while level 3 reaches 929 kB,
compared to LZO's 1581 kB.

However, when testing a full compilation of a large-ish source code (not
including linking), we find that CPU time is the bottleneck, shared
between the compressor and the preprocessors. The results of a best-of-5
run on this source are:

	lzo:    22.989s
	zstd1:  20.421s
	zstd2:  20.554s
	zstd3:  20.630s
	zstd4:  20.762s
	zstd5:  21.355s
	zstd6:  21.341s

Therefore, this commit chooses level 1 as the default. The 11% gain in
build time is caused the reduction in both CPU contention and the
network transmission -- zstd level 1 compresses the above source to 1029
kB, which is almost 35% smaller than the LZO content.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -41,7 +41,7 @@ matrix:
   - os: osx
     before_install:
       - brew update
-      - brew install lzo docbook2x gdb ccache
+      - brew install lzo zstd docbook2x gdb ccache
   - os: linux
     sudo: true # for setcap so we can run the tests in chroot.
     compiler: clang
@@ -99,6 +99,7 @@ addons:
     - libcap-ng-dev
     - libcap-ng-utils
     - liblzo2-dev
+    - libzstd1-dev
     - docbook2x
     - realpath
     - gdb
--- a/.travis.yml
+++ b/.travis.yml
@@ -33,11 +33,11 @@ matrix:
   - os: linux
     sudo: true # for setcap so we can run the tests in chroot.
     compiler: gcc
-    dist: trusty
+    dist: xenial
   - os: linux
     sudo: true # for setcap so we can run the tests in chroot.
     compiler: clang
-    dist: trusty
+    dist: xenial
   - os: osx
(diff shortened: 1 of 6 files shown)


thiagomacieira:
ping

HenryMiller1:
why is this combining switching the travis host OS with Zstandard compression?  These seem like separate concerns.

thiagomacieira:
Because the ancient Ubuntu has no Zstd package, but the not-so-ancient one does.

thiagomacieira:
Then you'll either need to install libzstd yourself or you won't be able to upgrade to the next version of Icecream.

thiagomacieira:
ping

thiagomacieira:
re-ping

HenryMiller1:
I hate to loose the sanitizers.

thiagomacieira:
re-re-ping. The loss of the sanitisers is because Travis doesn't have a good OS set up for us to test with. Using a container of a different Linux distro that has more sensible is probably recommended, but out of scope for me.

thiagomacieira:
I guess you're not interested.

llunak:
This feature is too good to be dropped just because Travis uses broken/ancient build setup. I've picked this for request #465 and then hacked the Travis setup into getting both sanitizers and zstd working somehow.

PR #433, opened 2018-11-20
Add support for packaging the environment as .tar.xz and .tar.zst

thiagomacieira:
In my setup, the file sizes are:
  .tar.gz:  193519 kB (default compression)
  .tar.gz:  192711 kB (level 9 compression)
  .tar.zst: 122330 kB
  .tar.xz:  109261 kB

The default compression level for xz (-6) is sufficient: it's fast
enough and compresses well enough. Using -9 gains 2.3% in size at the
expense of taking nearly 6x longer.

### DIFF
--- a/client/remote.cpp
+++ b/client/remote.cpp
@@ -164,7 +164,7 @@ rip_out_paths(const Environments &envs, map<string, string> &version_map, map<st
 
     Environments env2;
 
-    static const char *suffs[] = { ".tar.bz2", ".tar.gz", ".tar", ".tgz", NULL };
+    static const char *suffs[] = { ".tar.xz", ".tar.zst", ".tar.bz2", ".tar.gz", ".tar", ".tgz", NULL };
 
     string versfile;
 
--- a/daemon/environment.cpp
+++ b/daemon/environment.cpp
@@ -453,13 +453,17 @@ pid_t start_install_environment(const std::string &basename, const std::string &
     }
 
     fmsg = dynamic_cast<FileChunkMsg*>(msg);
-    enum { BZip2, Gzip, None} compression = None;
+    const char *decompressor = NULL;
 
     if (fmsg->len > 2) {
         if (fmsg->buffer[0] == 037 && fmsg->buffer[1] == 0213) {
-            compression = Gzip;
+            decompressor = "-z";    // --gzip
         } else if (fmsg->buffer[0] == 'B' && fmsg->buffer[1] == 'Z') {
-            compression = BZip2;
+            decompressor = "-j";    // --bzip2
+        } else if (fmsg->buffer[0] == 0xfd && fmsg->buffer[1] == 0x37) {
+            decompressor = "-J";    // --xz
+        } else if (fmsg->buffer[0] == 0x28 && fmsg->buffer[1] == 0xb5) {
+            decompressor = "-Iunzstd";
         }
     }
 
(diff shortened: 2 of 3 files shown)


thiagomacieira:
Notes/questions:
1. Should I split this into two patches, one adding support for decompressing .xz and .zst environments, and another for creating them?
1. Do we want to automatically create .xz? That would be incompatible with the existing protocol, since the running iceccd wouldn't be able to decompress them.

ossilator:
ps: speaking of heterogeneous environments, we have nodes go haywire all the time, so i suspect there is a subtle incompatibility somewhere, or maybe just some neglected error handling. it would require some real dedication to figure that out ...

thiagomacieira:
Well, I can split this and I think that the decompressor change makes sense. For the compressor, I'll add a option to choose which one you want. Maybe in the future we'll be able to make xz or zstd it the default.

ossilator:
an option makes no sense in a mixed environment, beyond just disabling it for everyone, and that really doesn't seem like a terribly desirable state.

thiagomacieira:
Right, it doesn't make sense in a mixed environment. But if you have a homogeneous environment, it might.

ossilator:
what for? when everyone has a properly built new version, then there is no practical advantage to turning it off. and otherwise it's a mixed env again.

thiagomacieira:
Reducing the network traffic and caches on the servers. A 40% reduction is not negligible.

By the way, one thing I didn't find while looking at the source: how often does iceccd unpack the environment for a given target? Once per job? Then zstd makes sense since it decompresses a lot faster.

ossilator:
i asked what sense it makes to turn *off* the feature when it is consistently available. i don't see any relevant downsides to it (yes, packing the environment takes a few millisecs more - that's not relevant).

i don't know how the env ends up in each chroot. @llunak will be able to tell.

coolo:
how often it uncompresses, depends on how many active environments you have in your network. It saves 100MB worth of (unpacked) environments for reuse.

What I wonder: how can you expect zstd to be available on servers just because it's available on clients?

coolo:
Basically: I don't think it's worth the network savings to risk not being able to unpack the environment. And servers don't store compressed environments, so there is no real storage gain.

HenryMiller1:
Do we know that all systems support xz?  BSD?  Does tar take a different options on those systems?

thiagomacieira:
Please look at the updated change. Despite the name of the branch, the commit is no longer trying to compress the environment as xz. It's just updating the decompressor so it can cope with xz and zst compressed tarballs, if someone sends them.

thiagomacieira:
I don't know how to do that type of feature negotiation. `IS_PROTOCOL_40` might suffice, for the sender side.

RedBeard0531:
Since you are already making icecreamd zstd-aware, could you just have it handle decompressing the environment before handing it off to tar? Then you don't need to care if the host is able to untar .zst files because the host will never see them.

thiagomacieira:
It could be done, but is it worth it? Two reasons:

1. because we'll have to force people to have libzstd installed, chances are zstd tool will also be
2. since we won't be able to actually start creating .tar.zst or .tar.xz any time soon for compatibility reasons, by the time we can, having zstd tool should be common place

RedBeard0531:
I'm not sure I agree.

1. If you are a building icecream for anything other than packaging in a distro's repo, you could (should?) statically link in libzstd so you don't need to depend on it being in some ancient distro's repos. This would also let you support a newer (and hopefully better/faster) version of zstd than would be present on that system. In this case, there is no reason to expect the zstd binary to be installed, and also no reason to expect the system tar to support decompressing .tar.zst files.
2. At least for us, it is the build system's job to [create the icecream environment](https://github.com/RedBeard0531/mongo_module_ninja/blob/965782ecf223073b47ce6fcebb73132d497194b0/build.py#L197-L264) as part of the normal build process. We could easily make it produce .tar.zst files if zstd is on the path (or just bundle a copy of zstd with the build system so it is always available) even if the default `icecc_create_env` script can't. Then we could just say that if you want faster builds, put zstd on your path, but if we require everyone in the cluster to use an icecc with zstd support, it won't need to have zstd installed on the *remote* host to work. This would also allow a more gradual migration if jobs with a zstd env forced a newer protocol version, they would only be sent to remote hosts that are new enough to be zstd enabled. The key benefit is that you would only need to install a new iceccd, not a whole new distro to take advantage of this.

thiagomacieira:
I don't see why you should statically link. If you build libzstd with a simple `configure && make install`, the library will be in /usr/local and so will the executable. And if you can force everyone in your cluster to upgrade to a version that was linked to libzstd, then you can ask them to put zstd on PATH too.

thiagomacieira:
ping

PR #439, opened 2018-12-12
Use Linux's unshare(2) abilities to further limit what compilers can do

thiagomacieira:
It's the same principle as containers. We limit:
- CLONE_NEWIPC (SysV IPC): accessing IPC in the system
- CLONE_NEWNET (networking): network access (only loopback allowed)
- CLONE_NEWNS (mount namespace): complements chroot
- CLONE_NEWPID (PIDs): no ptrace(2) or kill(2) any other processes
- CLONE_NEWUSER (users): no other processes share UID
- CLONE_NEWUTS (UTS)

As a side-effect of CLONE_NEWPID, when run, the compiler will see itself
as PID 1.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -108,6 +108,7 @@ AC_CHECK_HEADERS([resolv.h], [], [],
 # include <arpa/nameser.h>
 #endif
 ])
+AC_CHECK_HEADERS([sched.h], [AC_CHECK_FUNCS([unshare])])
 
 AC_ARG_VAR(TAR, [Specifies tar path])
 AC_PATH_PROG(TAR, [tar])
--- a/daemon/environment.cpp
+++ b/daemon/environment.cpp
@@ -33,6 +33,9 @@
 #include <sys/types.h>
 #include <sys/wait.h>
 #include <signal.h>
+#ifdef HAVE_SCHED_H
+#include <sched.h>
+#endif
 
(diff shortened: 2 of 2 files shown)


thiagomacieira:
Can't reproduce the failure on my openSUSE Tumbleweed.

thiagomacieira:
ping. Please advise what to do.

llunak:
I can reproduce that on openSUSE 15.0/1 . Both Clang and GCC get stuck in read(), presumably trying to read stdin data that never comes, because the local cpp has failed. I don't understand that, because stdin should get EOF, and the flag that triggers this is CLONE_NEWPID.

HenryMiller1:
I'd like to have this (ideally on more than linux), but the hang is concerning.  I added help wanted, it is probably a hard task to figure out why the hang happens on some system.  If we can figure out why it hangs in some cases we can figure out what to do.

HenryMiller1:
On thing that comes to mind is perhaps old versions are buggy.  If we can figure out where the bug was fixed we can disable this.  If the bug is old enough we can document those systems as unsupported.  It is worth rebasing with the latest where we have different CI systems.

deriamis:
I'm going to take a look at this in the coming couple of weeks and see if I can figure out what's going on here.

PR #443, opened 2019-01-10
use tar from PATH instead of configure time absolute path

lynxeye-dev:
This makes the tar invocation robust against changed tar paths between build and execution host.


HenryMiller1:
We are using libarchive now which seems like a better approach long term.  Thanks for the attempt.

Issue #414, opened 2018-08-24
__COUNTER__ expanded inside directive with -fdirectives-only


lynxeye-dev:
Since the switch to use -fdirectives-only on GCC I'm no longer able to build some packages. One such example is GNU M4, which fails with the error message in the title. This seems to be a corner case where -fdirectives-only can not be used [1]. As this error happens while running the local CPP half for a remote build it is treated as a fatal error, failing the build.

It seems there is no other way to fix this than to detect this error message from the local CPP and fall back to a local build. As this requires more changes to the icecc client than I currently have time to do, I'm documenting the issue here. Maybe someone other can help to fix this.

https://gcc.gnu.org/ml/gcc-patches/2007-06/txt00139.txt

stayprivate:
googlebenchmark also uses __COUNTER__ and creates issue. However it's usage is somewhat optional hence a small change in googlebenchmark header file will work around the issue.

LeSpocky:
FWIW, here's the output of a failing build of m4 1.4.18 with DistroKit and [ptxdist](https://www.ptxdist.org/) and [ICECC 1.2.0 from Debian 10](https://packages.debian.org/buster/icecc):

https://hastebin.com/lirinilaqo.js

@rohieb could reproduce this.

rohieb:
Just for googlability, I also got the same error with libtasn, gettext, glib and gdbserver, which seem to use the same `verify.h` with the `__COUNTER__` definition like m4.

HenryMiller1:
Merged proposed fix

PR #444, opened 2019-01-10
try local recompile when local cpp invocation fails

lynxeye-dev:
Proposed fix for #414.


rohieb:
Hmm. Travis failures seem to be unrelated to the patch?

HenryMiller1:
Can you rebase this on top to the latest and force push?  That will re-run the travis build, and also run additional CI jobs that have been added since.

HenryMiller1:
The code looks good, and I agree it looks like a CI issue, but I reflexively don't like to look at code until CI passes.

lynxeye-dev:
I just pushed an update which fixes all the test fallout caused by the local recompile fallback.

PR #447, opened 2019-02-21
Removed dependency on native tar command and used libarchive library

albuquerque:
Replaced call to tar utility and added support for libarchive library instead as we were having issues with tar hanging up infinitely for EOF forcing us to restart iceccd. This gives the ability to use all environments supported by libarchive without having to depend on an external utility.

Also added more descriptive log messages.

### DIFF
--- a/daemon/environment.cpp
+++ b/daemon/environment.cpp
@@ -112,7 +112,7 @@ static bool exec_and_wait(const char *const argv[])
         // parent
         int status;
 
-        while (waitpid(pid, &status, 0) < 0 && errno == EINTR);
+        while (waitpid(pid, &status, 0) < 0 && errno == EINTR) {}
 
         return shell_exit_status(status) == 0;
     }
@@ -401,7 +401,7 @@ size_t finish_create_env(int pipe, const string &basedir, string &native_environ
     char buf[1024];
     buf[0] = '\0';
 
-    while (read(pipe, buf, 1023) < 0 && errno == EINTR);
+    while (read(pipe, buf, 1023) < 0 && errno == EINTR) {}
 
     if (char *nl = strchr(buf, '\n')) {
         *nl = '\0';
@@ -458,7 +458,7 @@ pid_t start_install_environment(const std::string &basename, const std::string &
                                 int &pipe_to_stdin, FileChunkMsg *&fmsg,
                                 uid_t user_uid, gid_t user_gid, int extract_priority)
 {
-    log_error() << "start_install_environment: " << basename << " target "<<target << " Name: " << name << endl;
+    log_info() << "start_install_environment: " << basename << " target "<<target << " Name: " << name << endl;
     if (!name.size()) {
         log_error() << "illegal name for environment " << name << endl;
         return 0;
@@ -593,7 +593,7 @@ pid_t start_install_environment(const std::string &basename, const std::string &
     archive_write_disk_set_standard_lookup(ext);
 
     if(archive_read_open_fd(a, 0, fmsg->len)){
-        trace() << "start_install_environment: "<< "archive_read_open_fd faied"<< endl;
+        log_error() << "start_install_environment: "<< "archive_read_open_fd faied"<< endl;
         _exit(1);
     }
 
@@ -604,7 +604,7 @@ pid_t start_install_environment(const std::string &basename, const std::string &
             break;
         }
         if (r < ARCHIVE_WARN){
-            trace() << "start_install_environment: r  < ARCHIVE_WARN " <<archive_error_string(a)<<endl;   
+            log_error() << "start_install_environment: r  < ARCHIVE_WARN " <<archive_error_string(a)<<endl;   
             _exit(1);
         }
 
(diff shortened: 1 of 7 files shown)


daemon/environment.cpp
```diff
@@ -109,7 +112,7 @@ static bool exec_and_wait(const char *const argv[])
         // parent
         int status;
 
-        while (waitpid(pid, &status, 0) < 0 && errno == EINTR) {}
```
HenryMiller1: I like the old way better.  They are equivalent, but the {} makes it clear that the loop doing nothing is intentional.

daemon/environment.cpp
```diff
@@ -398,7 +401,7 @@ size_t finish_create_env(int pipe, const string &basedir, string &native_environ
     char buf[1024];
     buf[0] = '\0';
 
-    while (read(pipe, buf, 1023) < 0 && errno == EINTR) {}
+    while (read(pipe, buf, 1023) < 0 && errno == EINTR);
```
HenryMiller1: As above, {} is clearer in intent.

daemon/environment.cpp
```diff
@@ -425,11 +428,37 @@ size_t finish_create_env(int pipe, const string &basedir, string &native_environ
...
+        }
+    }
+}
+
 pid_t start_install_environment(const std::string &basename, const std::string &target,
                                 const std::string &name, MsgChannel *c,
                                 int &pipe_to_stdin, FileChunkMsg *&fmsg,
                                 uid_t user_uid, gid_t user_gid, int extract_priority)
 {
+    log_error() << "start_install_environment: " << basename << " target "<<target << " Name: " << name << endl;
```
HenryMiller1: log_error is the wrong level for this. You want log_info

daemon/environment.cpp
```diff
@@ -540,32 +557,81 @@ pid_t start_install_environment(const std::string &basename, const std::string &
...
+
+    a=archive_read_new();
+    archive_read_support_format_all(a);
+    archive_read_support_filter_all(a);
+    ext = archive_write_disk_new();
+    archive_write_disk_set_options(ext, flags);
+    archive_write_disk_set_standard_lookup(ext);
+
+    if(archive_read_open_fd(a, 0, fmsg->len)){
+        trace() << "start_install_environment: "<< "archive_read_open_fd faied"<< endl;
```
HenryMiller1: This should be log_error

daemon/environment.cpp
```diff
@@ -540,32 +557,81 @@ pid_t start_install_environment(const std::string &basename, const std::string &
...
+    }
+
+    for(;;){
+        int r = archive_read_next_header(a, &entry);
+        if (r == ARCHIVE_EOF) {
+            trace() << "start_install_environment: reached end of archive while recveiving environment "<< endl;
+            break;
+        }
+        if (r < ARCHIVE_WARN){
+            trace() << "start_install_environment: r  < ARCHIVE_WARN " <<archive_error_string(a)<<endl;
```
HenryMiller1: log_error

daemon/environment.cpp
```diff
@@ -540,32 +557,81 @@ pid_t start_install_environment(const std::string &basename, const std::string &
...
+        /*Extracting archive*/
+        const char* currentFile = archive_entry_pathname(entry);
+        const std::string fullOutputPath = dirname + "/"+currentFile;
+        archive_entry_set_pathname(entry, fullOutputPath.c_str());
+        r = archive_write_header(ext, entry);
+
+        if(archive_entry_size(entry) > 0){
+            r= copy_data(a, ext);
+            if(r < ARCHIVE_WARN){
+                trace()<< "start_install_environment: " << archive_error_string(ext)<<endl;
```
HenryMiller1: isn't this an error - use log_error

daemon/environment.cpp
```diff
@@ -540,32 +557,81 @@ pid_t start_install_environment(const std::string &basename, const std::string &
...
+                _exit(1);
+            }
+        }
+        r=archive_write_finish_entry(ext);
+        if(r<ARCHIVE_WARN){
+            _exit(1);
```
HenryMiller1: shouldn't we log an error here?

daemon/environment.cpp
```diff
@@ -614,7 +678,7 @@ size_t remove_environment(const string &basename, const string &env)
     if (pid) {
         int status = 0;
 
-        while (waitpid(pid, &status, 0) < 0 && errno == EINTR) {}
```
HenryMiller1: I prefer {}

daemon/environment.cpp
albuquerque: done.

daemon/environment.cpp
albuquerque: done.

daemon/environment.cpp
albuquerque: done

daemon/environment.cpp
albuquerque: done

daemon/environment.cpp
albuquerque: done

albuquerque:
As requested, I have updated the NEWS and README files as well in the latest commit.

HenryMiller1:
Next release will probably be 1.3, but that can be considered latter.

HenryMiller1:
Thanks for your work.

PR #465, opened 2019-06-01
Add support for Zstandard compression of the content (#2)

llunak:
This includes the commit from request #432 and then another one that beats Travis into getting both zstd and sanitizers working.


thiagomacieira:
Thank you, Luboš (and LTNS).

HenryMiller1:
Thanks for the work, it was well worth doing.

Issue #466, opened 2019-06-04
expose `min_mem_limit` to command lni


timblechmann:
the hardcoded `min_mem_limit` seems to be very low for c++ sources and i tend to have remote compile failures for c++ sources despite having plenty of memory. it would be nice if the hardcoded limit of 100MB could be increased or exposed to the user

timblechmann:
hmm, actually i'm wondering if there's something wrong with the free memory detection on macos: it seems to constantly set the limit to 100mb ... on a 16GB agent

PR #467, opened 2019-06-04
daemon: fix `MemTotal` on macos

timblechmann:
the mach codepath doesn't obtain `MemTotal`, which will always cause the
a pessimistic 100MB memory limit. as later on, we'll just return 0.

the best way on macos to obtain the total memory is reading `hw.memsize`

### DIFF
--- a/daemon/load.cpp
+++ b/daemon/load.cpp
@@ -54,6 +54,10 @@
 #include <devstat.h>
 #endif
 
+#ifdef __APPLE__
+#include <sys/sysctl.h>
+#endif
+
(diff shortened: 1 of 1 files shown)


timblechmann:
compare: #466

zbeekman:
Can someone ping me when/if this is merged? I'd like to backport this patch to the homebrew formula.

HenryMiller1:
@zbeekman merged.

timblechmann:
@HenryMiller1 thanks for merging it so quickly!

PR #468, opened 2019-06-06
handle -MF, -MT and -MQ when passed via -Wp

lynxeye-dev:
More fixes for Linux kernel build with -fdirectives-only. Some of the
kernel tools are passing the above options directly to the preprocessor
via -Wp. Need to make sure they are applied on the local cpp side.


llunak:
Why did you use str_startswith() for the new checks instead of strcmp() as the old checks do?

PR #472, opened 2019-06-13
Bundle all internal clang includes in toolchain

### LINKED ISSUE #471: macos: compile failures without `ICECC_CLANG_REMOTE_CPP=0`
(apple)clang seems to require local preprocessing, otherwise i'm getting compile errors (e.g. `INT_MAX` and friends are undefined).

would it make sense to use local preprocessing by default on macos, considering that most FAQs on icecream/macos seem to suggest it?

i'm happy to provide a patch and backport it into homebrew

torarnv:
Fixes #471

### DIFF
--- a/client/icecc-create-env.in
+++ b/client/icecc-create-env.in
@@ -405,13 +405,13 @@ if test -n "$clang"; then
     fi
 
     # clang always uses its internal .h files
-    clangincludes=$($orig_clang -print-file-name=include/limits.h)
-    if test -z "$clangincludes"; then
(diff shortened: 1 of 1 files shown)


torarnv:
It's still not clear to me _why_ we need to bundle the includes, but if we need to we should bundle all of them.

torarnv:
This is a bit too much on linux, as it results in bundling all of `/usr/include`. Will look into it

torarnv:
#473 is the right way to go

PR #473, opened 2019-06-16
do not package Clang internal headers in icecc-create-env

llunak:
It doesn't seem to be actually needed (tested with Clang 8.0.0 and 4.0.1).
I have no idea why I added it, the commit doesn't say. Possibly
initially -frewrite-includes had a bug that necessitated this.

### DIFF
--- a/client/icecc-create-env.in
+++ b/client/icecc-create-env.in
@@ -403,17 +403,6 @@ if test -n "$clang"; then
         touch $tempdir/fakeproc/proc/cpuinfo
         add_file $tempdir/fakeproc/proc/cpuinfo /proc/cpuinfo
     fi
-
-    # clang always uses its internal .h files
-    clangincludes=$($orig_clang -print-file-name=include/limits.h)
-    if test -z "$clangincludes"; then
-        echo $orig_clang cannot find its includes
-        exit 1
-    fi
-    clangincludes=$(dirname $(abs_path $clangincludes))
-    for file in $(find $clangincludes -type f); do
-      add_file "$file"
-    done
 fi
 
 # Do not do any prefix stripping on extra files, they (e.g. clang plugins) are usually


torarnv:
Looks good to me :shipit: 

Qt builds fine with this minimal toolchain as long as I explicitly include the needed headers as described in #471

alcroito:
Minor note:
When i used Xcode 10.1 as the source for all relevant headers, and sent cross compiling jobs on a Linux host, targetting darwin, using the clang compiler tagged with swift 5.0.1 (llvm 7)  from https://hub.docker.com/_/swift , I got a lot of build failures not finding intrinsics.

This is using the environment created by icecc-create-env from this PR. So clang headers where not packaged in the enviornment.

The issue disappeared when i used the same icecc environment, but updated my xcode to 10.2.

I don't know if this is a discrepancy in the Xcode version <-> swift-provided clang compiler, or something to do with the older version of Xcode.

So mostly just some fear of breaking icecc-create-env when using older xcode versions.

llunak:
Can you be more specific about the problems? E.g. provide a specific error this triggers?

I can scratch this change if it turns out it is needed after all, so I'd like to investigate this more.

alcroito:
Here's the build log of the failure while building Qt 5.14 qtbase repo.
https://paste.ubuntu.com/p/MvPTXNH9TJ/

The circumstances:
I'm calling make on macOS 10.14, with Xcode 10.1 used while configuring.
On the same machine I run
export ICECC_VERSION=x86_64:/Volumes/T3/Dev/docker_icecream_envs/swift-llvm-5.0.1-x86_64.tar.gz

to use just the cross compiler environment which was built using https://github.com/torarnv/icecc-create-x86_64-env and your PR. So the cross compiler is ran on various Linux machines in our network.

Keeping everything the same, but selecting Xcode 10.2.1 instead of 10.1, then the build succeeds.
Not that the cross compiler environment has no includes bundled.

torarnv:
> My suspicion is that it could be a discrepancy between Xcode 10.1 and swift-llvm-5.0.1.
> The equivalent swift llvm version for Xcode 10.2.1 is 5.0.1.
> The equivalent swift llvm version for Xcode 10.1 is 4.2.1.

Yes, I don't think those issues were due to the lack of headers in the env. You can probably verify this by doing the same setup but with the headers present, i.e. Xcode 10.1  + swift-llvm-5.0.1-based env _with_ headers.

llunak:
Can you please confirm whether the problem goes away if you use icecc-create-env without this change? It is not clear from your description, you only mention the problem going away when switching to 10.2, but that's irrelevant here.

alcroito:
Without the PR, Xcode 10.1, swift-llvm-5.0.1 with intrinsic headers, it still fails with the same errors not finding intrinsic functions. So I guess false alarm from my side.

PR #490, opened 2019-07-17
icecc-create-env: ensure that target directories of symbolic links get created

### LINKED ISSUE #464: icecc-create-env failure
We are struggling to create the env for icecream on fedora (29 & 30 at least). Line 453 of script icecc-create-env, there are these lines:
```
# special case for weird multilib setups
for dir in /lib /lib64 /usr/lib /usr/lib64; do
    test -L $dir && cp -Pp $dir $tempdir$dir
done
```
On fedora, /lib64 is a symlink, so the command cp -Pp copy the symlink without the content, (the cp -Pp was introduced in commit 67d989c4ee8d9daea09e5855d2312c6aec706784 six months ago). 
But line 470, the `mkdir -p $tempdir/$(dirname $target)` command failed, because you can't create a folder if a symlink with the same name already exist, so we have the following error :
```
mkdir: impossible de créer le répertoire « /tmp/iceccenvwKe7Iy//lib64 »: File exists
```
and `cp` line 471 failed too, making the creation of the archive impossible. 
For us, simply commenting lines 454,455,456 seems to works. Any thoughts on this ?

mp-chet:
This is an attempt to fix #464. I am not 100 % certain that this is the correct way to tackle the issue, so I'm open for constructive feedback.

### DIFF
--- a/client/icecc-create-env.in
+++ b/client/icecc-create-env.in
@@ -527,7 +527,10 @@ fi
 
 # special case for weird multilib setups
 for dir in /lib /lib64 /usr/lib /usr/lib64; do
-    test -L $dir && cp -Pp $dir $tempdir$dir
+    if [ -L $dir ]; then
+        cp -Pp $dir $tempdir$dir
+        mkdir -p $tempdir/$(file $dir | cut -d' ' -f5-)
+    fi
 done
 
 new_target_files=


llunak:
Based on discussion in #464, I've simply removed that code.

PR #491, opened 2019-07-18
 Improve error messages for exec() calls 

mp-chet:
If `exec()` fails, the user should be told which binary was supposed to be
executed.

### DIFF
--- a/client/cpp.cpp
+++ b/client/cpp.cpp
@@ -200,6 +200,8 @@ pid_t call_cpp(CompileJob &job, int fdwrite, int fdread)
     dcc_increment_safeguard(SafeguardStepCompiler);
     execv(argv[0], argv);
     int exitcode = ( errno == ENOENT ? 127 : 126 );
-    log_perror("execv failed");
+    ostringstream errmsg;
+    errmsg << "execv " << argv[0] << " failed";
+    log_perror(errmsg.str());
     _exit(exitcode);
 }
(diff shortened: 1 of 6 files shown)


daemon/environment.cpp
```diff
@@ -763,7 +763,7 @@ bool verify_env(MsgChannel *client, const string &basedir, const string &target,
     reset_debug();
     chdir_to_environment(client, dirname, user_uid, user_gid);
     execl("bin/true", "bin/true", (void*)NULL);
-    log_perror("execl failed");
+    log_perror("bin/true");
```
llunak: This will lead to error message "bin/true(Error: No such file or directory)", which is confusing. If you think the binary to execute is important there, please add it to the existing error message.

mp-chet: Done.

llunak: I commented on this last one, but I meant it as a generic comment that affects all the changes. It should be something like log_perror("execl failed (" + argv[ 0 ] + ")"), with the necessary std::string handling.

mp-chet: Okay. I'm using `std::ostringstream` for concatenating the strings.

client/cpp.cpp
```diff
@@ -200,6 +200,8 @@ pid_t call_cpp(CompileJob &job, int fdwrite, int fdread)
     dcc_increment_safeguard(SafeguardStepCompiler);
     execv(argv[0], argv);
     int exitcode = ( errno == ENOENT ? 127 : 126 );
-    log_perror("execv failed");
+    ostringstream errmsg;
+    errmsg << "execv " << argv[0] << " failed";
```
llunak: I myself would have written string( "execv" ) + argv[0] + "failed", but fair enough.

llunak:
Ok, thank you.

PR #496, opened 2019-08-12
In icecc-create-env's --addfile argument, allow specifying a target path

smokris:
In addition to the files that `icecc-create-env` automatically finds, I'd like to be able to add my own files at specific destination paths, but currently `icecc-create-env` assumes that the source path should equal the destination path.  This patch allows changing the destination path by adding `=targetpath` to the end of each `--addfile` argument.

### DIFF
--- a/client/icecc-create-env.in
+++ b/client/icecc-create-env.in
@@ -22,7 +22,7 @@ usage ()
     echo "Usage: $0 <compiler_binary> [extra_options]"
     echo "For GCC, pass the the gcc binary, the matching g++ will be used automatically."
     echo "For Clang, pass the clang binary."
-    echo "Use --addfile <file> to add extra files."
+    echo "Use --addfile <file>[=<targetpath>] to add extra files."
     echo "Use --compression <type> to set tarball type (none,gzip,bzip2,zstd,xz)."
     echo "For backwards compatibility, the following is also supported:"
     echo "$0 --gcc <gcc_path> <g++_path>"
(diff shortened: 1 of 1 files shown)


llunak:
What is the use case for this?

smokris:
> [@llunak] What is the use case for this?

I'm trying to use Icecream on a project that uses a custom build of LLVM/Clang to create a custom compiler.

- The project uses [Conan](https://conan.io) to install a custom LLVM/Clang build.  I first tried this:

      icecc-create-env ~/.conan/data/llvm/myversion/myproject/stable/package/mybuildhash/bin/clang++

   …but compilation using that environment fails because the compiler relies on shared libraries in `~/.conan/data/llvm/myversion/myproject/stable/package/mybuildhash/lib`, which `icecc-create-env` didn't automatically find.

   If I add them using e.g. `--addfile ~/.conan/data/llvm/myversion/myproject/stable/package/mybuildhash/libLLVMAnalysis.dylib`, then they get placed in `/home/me/.conan/…`.  The compiler references them using a relative path (`@loader_path/../lib/libLLVMAnalysis.dylib`) so it can't find them at that absolute path since `icecc-create-env` has relocated the compiler to `/usr/bin`.

   With this patch, I can specify that I'd like to place the shared libraries in `/usr/lib`, and then the compiler is able to find them and distributed compilation succeeds.

- Secondly, after the project finishes building its custom compiler, I'd like to use Icecream to distribute invocations of that custom compiler.  The custom compiler relies on some shared libraries and other resources, which I'd like to be able to include at specific paths in the compiler environment.

llunak:
The icecc-create-env script is supposed to package the compiler properly, so if that doesn't work here, the script should get adapted to handle your case. I consider this change to be just a workaround.

There is a "stripprefix" variable in the script that is supposed to strip all such prefixes when relocating the compiler. It is set to "stripprefix=$(dirname $(dirname $added_clang))", which seems it should work for your case, so you should check why it doesn't.

llunak:
Based on my previous comment, I consider your case to be a bug (or possibly a misconfiguration) and this change would be just a workaround. So closing this. If you have figured out why the prefix is not stripped properly, please submit a fix for that.

Issue #509, opened 2019-11-14
[macos] icecc-create-env misses some dylibs


timblechmann:
on macos 10.14.6 (xcode 10 and 11), it seems that `icecc-create-env` is missing to bundle come dylibs from the system. the iceccd log on the agent contains:

```
[66472] 2019-11-14 02:43:17: compiler produced stderr output:
dyld: Library not loaded: /usr/lib/system/libcache.dylib
  Referenced from: /usr/lib/libSystem.B.dylib
  Reason: image not found
```

i've started bundling (`--addfile`) one dylib after the other from `/usr/lib/system` and eventually from `/usr/lib/`, but ended up bundling *all* dylibs from those folders in order to get a working build environment (after adding 10ish dylibs manually, i took the shortcut of bundling everything).

nevertheless, i wonder why those dylibs haven't been found automatically, neither when bundling the trampoline executable `/usr/bin/clang`, nor the `clang` binary inside xcode.

permcody:
We are having the same issue. We verified that the two tagged releases before 1.3 work, but 1.3 doesn't seem to work on Mac OS. We've tried different combinations of XCode and Command Line Tools to no avail. Definitely something up with the latest release. 

More info: We have a big Mac installation at our site. We have around 20-25 Macs on the Icecream network running a mix of MacOS versions. 1.1 is pretty solid as we've been using it for a few years. We'd like to move to 1.3 to get some of the better "downed client" and "blacklist" scheduler features.

Thanks for the information @timblechmann.

permcody:
UPDATE: The problem is the `icecc-create-env` tool itself. We discovered that if we use an environment built with the 1.1 release (using `icecc-create-env`), it works properly using the latest release. 

`icecc-create-env` is broken on MacOS

HenryMiller1:
The main developers of icecc are not macos experts, I don't think any of us even have a Mac to work with. We would encourage one of you to figure this out, it shouldn't be that hard. Only a few changes have been made to icecc-create-env since 1.1 so you should be able to find the offending change quickly and then make a fix that works.

Harder but worth also doing would be get some tests working on the CI system. Right now we have no idea if we have broke macos. If it fails to build we will notice and fix it, but if it builds correctly we assume it will work (which obviously is not a correct)

loganharbour:
The offending commit for our environment is 3ac9141.

I'll take a look when I have a moment.

permcody:
[...]

PR #511, opened 2019-11-20
Revert Darwin environment creation optimization

loganharbour:
Reverts the changes for OS X in 3ac9141, which led to missing dylibs. This doesn't correct the optimization, it removes it entirely so that it works for now.

See #509.


llunak:
Thank you.

PR #515, opened 2020-02-05
The tilde doesn't need to be escaped in markdown.

albert-github:
- The tilde doesn't need to be escaped in markdown.
- placing some strings into "code" format by means of backticks.


llunak:
Thank you.

PR #518, opened 2020-02-26
Compile locally with coverage argument

efokin:
According
https://gcc.gnu.org/onlinedocs/gcc/Instrumentation-Options.html
--coverage option is a synonym for -fprofile-arcs -ftest-coverage.


llunak:
Thank you.

Issue #527, opened 2020-04-12
flush_writebuf() failed


zub2:
Running the latest icecc (current master - 16d7b3a717) I'm frequently seeing errors complaining about flush_writebuf failed. I was able to reproduce this with just one scheduler and two iceccd daemons. In my test case the remote daemon accepts 12 jobs and I'm building a fairly large C++ project. The errors printed by the icecc compiler wrapper look like:
```
ICECC[31580] 2020-04-08 16:54:29: timed out while trying to send data
ICECC[31580] 2020-04-08 16:54:29: flush_writebuf() failed(Error: Resource temporarily unavailable)
ICECC[31580] 2020-04-08 16:54:29: write of source chunk to host 10.x.y.z
ICECC[31580] 2020-04-08 16:54:29: failed (Error: Resource temporarily unavailable)
ICECC[31580] 2020-04-08 16:54:29: got exception Error 15 - write to host failed (10.x.y.z) 
```
Digging deeper into the issue and adding some more traces, I get the following from the wrapper:
```
ICECC[4972] 2020-04-12 17:29:24: Have to use host 10.x.y.z:10245 - Job ID: 25472 - env: x86_64 - has env: true - match j: 0
ICECC[4972] 2020-04-12 17:29:26: connected to 10.x.y.z
ICECC[4972] 2020-04-12 17:29:28: <send compile_file>
ICECC[4972] 2020-04-12 17:29:28: </send compile_file: 0ms>
ICECC[4972] 2020-04-12 17:29:28: <build_remote_int[jobId=25472]: dcc_lock_host>
ICECC[4972] 2020-04-12 17:29:28: dcc_lock_host_slot: attempting to lock /tmp/.icecream-blah/local_lock, block=0
ICECC[4972] 2020-04-12 17:29:28: dcc_lock_host_slot: attempting to lock /tmp/.icecream-blah/local_lock1, block=0
ICECC[4972] 2020-04-12 17:29:28: dcc_lock_host_slot: attempting to lock /tmp/.icecream-blah/local_lock2, block=0
ICECC[4972] 2020-04-12 17:29:28: dcc_lock_host_slot: attempting to lock /tmp/.icecream-blah/local_lock3, block=0
ICECC[4972] 2020-04-12 17:29:28: <dcc_lock_host_slot blocking (idx=0, max_cpu=4)>
[...]

PR #528, opened 2020-04-13
do not keep host lock for cpp when sending network data

llunak:
We keep host locks when invoking cpp to limit the number of invocations,
but cpp runs in parallel to sending its output over the network, and
that doesn't need to be limited this way. And moreover it causes
timeouts, see https://github.com/icecc/icecream/issues/527 .


llunak:
I've already accidentally pushed this in with other changes, closing.

PR #531, opened 2020-04-14
Read and save the available preprocessed input for the future child

joeyjyliu:
When the remote host is full with work, the wait time for it to free up and
fork a child to compile could be long. If the input is ready to read, we will read
them and save it for the future child; otherwise the write on the client side would be blocked.
As a result, the client would time out and give up.

### DIFF
--- a/daemon/main.cpp
+++ b/daemon/main.cpp
@@ -1978,11 +1978,14 @@ void Daemon::answer_client_requests()
         Client *client = clients.find_by_channel(c);
         assert(client);
         int current_status = client->status;
-        bool ignore_channel = current_status == Client::TOCOMPILE
-                              || current_status == Client::WAITFORCHILD
-                              || current_status == Client::WAITINSTALL;
+        bool ignore_channel = current_status == Client::WAITFORCHILD ||
+                              current_status == Client::WAITINSTALL;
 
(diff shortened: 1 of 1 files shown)


llunak:
What is the specific problem this is supposed to fix? It looks a bit like #527 to me, but that one has been just fixed in a different way.

joeyjyliu:
The error is similar (sending file timeout) but this one happens when the remote host is compiling fairly large files that would take one or two minutes to finish. The poll timeout on client is 30s so the client would time out in poll for writable socket. The socket at that time is full of file chunk messages. Usually as the remote host finishes the current compilation jobs it would fork (assuming there are not too many children already) and take more jobs and in the child it will continue to read file chunk messages thus unblock the client to write more. But in the case of long wait due to large compilation jobs and being limited by the max number of children on the daemon, the client would time out and give up. 

This patch is try to save the work the client has done on the daemon side and save it for the future child that daemon is going to create soon.

joeyjyliu:
What #527 fixed is the lockup on the client, while mine is on the lockup on the daemon.

joeyjyliu:
Ping. @llunak please kindly let me know if you need more input/clarifications. Thanks.

llunak:
Sorry, I needed time to have a deeper look at all the relevant state changes. The patch looks good, thank you.

PR #534, opened 2020-05-19
Use pkg-config to find packages

JPEWdev:
Package config should be used to find packages instead of manually
searching for headers. This makes cross-compiling easier, as well as
makes it possible to build a static binary (in particular, libarchive is
annoying to link statically without using pkg-config)


llunak:
Looks ok to me in general, but the Cirrus check fails, do you have any idea why? It built fine without your change.

JPEWdev:
> Looks ok to me in general, but the Cirrus check fails, do you have any idea why? It built fine without your change.

It looks like pkg-config can't find libarchive.pc.... I don't really know a lot about FreeBSD, but it appears that it can't find it for some reason. I'll going to try and boot a freebsd QEMU image to see if I can replicate it

JPEWdev:
> Looks ok to me in general, but the Cirrus check fails, do you have any idea why? It built fine without your change.

`libarchive` was missing from the package list

llunak:
Thank you.

PR #540, opened 2020-06-20
Don't explicitly forbid tunnels

torokati44:
This blocked me from adding a worker host to a scheduler through a WireGuard connection.
And since it was already in the TODO, and seemed like an easy fix... here we are!


torokati44:
I'm not sure if CI broke because some tests don't tolerate the new (and unexpected) packets coming through the tunnel interfaces of the testing rig? Or was it already broken?
I have no idea, but maybe someone should verify that it's not going to cause problems there...

llunak:
The tests broke because they cannot find a localhost interface for a broadcast, presumably your test changed the logic there. Run "make test" to reproduce the problem.

Issue #538, opened 2020-06-13
iceccd doesn't seem to obey --interface option


baybal:
Hello,

I'm running iceccs on a cluster where each machine is a member of the same public net, and a private net.

I'm trying to force iceccd to only look for private network for listening to scheduler server broadcasts, but it still tries to use interface connecting to a public net.

`root@SRVMT6CTXB6P5:~# iceccd -i ens192 -v --help
usage: iceccd [-n <netname>] [-m <max_processes>] [--no-remote] [-d|--daemonize] [-l logfile] [-s <schedulerhost[:port]>] [-v[v[v]]] [-u|--user-uid <user_uid>] [-b <env-basedir>] [--cache-limit <MB>] [-N <node_name>] [-i|--interface <net_interface>] [-p|--port <port>]
root@SRVMT6CTXB6P5:~# ip address show
1: lo: <LOOPBACK,UP,LOWER_UP> mtu 65536 qdisc noqueue state UNKNOWN group default qlen 1000
    link/loopback 00:00:00:00:00:00 brd 00:00:00:00:00:00
    inet 127.0.0.1/8 scope host lo
       valid_lft forever preferred_lft forever
    inet6 ::1/128 scope host 
       valid_lft forever preferred_lft forever
2: ens160: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1500 qdisc mq state UP group default qlen 1000
    link/ether 00:50:56:01:5e:0b brd ff:ff:ff:ff:ff:ff
    inet 178.170.221.41/24 brd 178.170.221.255 scope global ens160
       valid_lft forever preferred_lft forever
    inet6 fe80::250:56ff:fe01:5e0b/64 scope link 
       valid_lft forever preferred_lft forever
3: ens192: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1500 qdisc mq state UP group default qlen 1000
    link/ether 00:50:56:01:5e:0e brd ff:ff:ff:ff:ff:ff
    inet 10.0.0.24/24 brd 10.0.0.255 scope global dynamic ens192
       valid_lft 6589sec preferred_lft 6589sec
    inet6 fe80::250:56ff:fe01:5e0e/64 scope link 
       valid_lft forever preferred_lft forever
root@SRVMT6CTXB6P5:~# iceccd -i ens192 -vvv
[1677] 2020-06-13 16:50:41: ICECREAM daemon 1.3.1 starting up (nice level 5) 
[...]

PR #544, opened 2020-06-24
Consider the --interface argument when broadcasting for scheduler discovery

torokati44:
This PR aims to (at least partially) solve #538.
It does not add any filtering for incoming packets, but at least the daemon won't look for schedulers on any interface other than the one given as command line argument (if one is given...).
It might not be perfect yet, because the naming of the new variables might be confusing here and there, and the value of the `--interface` option has to be passed around a lot in all sorts of places, but I'm open to suggestions for improvement.
Maybe adding some comments would be enough to explain what the new variables/parameters are there for?
Also, this might break icemon.

### DIFF
--- a/daemon/main.cpp
+++ b/daemon/main.cpp
@@ -2227,7 +2227,7 @@ bool Daemon::reconnect()
 
     if (!discover || (nullptr == (scheduler = discover->try_get_scheduler()) && discover->timed_out())) {
         delete discover;
-        discover = new DiscoverSched(netname, max_scheduler_pong, schedname, scheduler_port);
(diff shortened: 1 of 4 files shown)


llunak:
The patch in principle looks good, but can you please make sure it's backwards source compatible (i.e. that icemon compiles without changes)? In some cases it's not necessary at all to pass the extra argument (e.g. DiscoverSched::sendSchedulerDiscovery() can use the member variable), in other cases it can be the last argument with a default value (which mostly will be also better API, with the argument not being very important in the argument list).

PR #545, opened 2020-07-03
Add `--serialize-diagnostics` as argument with space

cinakyn:
Only `-serialize-diagnostics` is included in the arguments with space.  
But `--serialize-diagnostics` is also valid argument as [document](https://clang.llvm.org/docs/genindex.html) says.  
This issue make the icecc client set invalid file as input.

 
image


llunak:
Thank you.

PR #546, opened 2020-07-04
Ignore -fdirectives-only and -frewrite-includes options

### LINKED ISSUE #451: GCC preprocessor with `-fdirectives-only` can break compilation with raw string literals
Example code:

```
int main()
{
    const auto vert = R"(
#hello)";
    return 0;
}
```

If you preprocess this with `gcc -E -fdirectives-only`, then it fails with:

```
test.cpp:4:2: error: invalid preprocessing directive #hello
 #hello)";
  ^~~~~
test.cpp:4:8: warning: missing terminating " character
 #hello)";
        ^
```

I'm reporting this here for others to find, but will will now also report a bug upstream for GCC. I don't know how icecream could solve this issue - it looks like a GCC bug to me.

deriamis:
Icecream is already managing these flags remotely, and they don't make sense for purely local builds. Therefore, Icecream should ignore these flags when they are specified.

This change fixes #451.

### DIFF
--- a/client/arg.cpp
+++ b/client/arg.cpp
@@ -709,6 +709,12 @@ bool analyse_argv(const char * const *argv, CompileJob &job, bool icerun, list<s
                 if (argv[i + 1]) {
                     args.append(argv[++i], Arg_Rest);
                 }
+            } else if (str_equal("-fdirectives-only", a)
+                       || str_equal("-frewrite-includes", a)) {
+                // These don't make sense to run locally, and we inject them where
+                // appropriate remotely if remote preprocessing is requested. Therefore,
+                // skip them if they were supplied.
+                ++i;
             } else {
                 args.append(a, Arg_Rest);
 


llunak:
What is the actual use case you're trying to solve? This doesn't help with #451, as that one is about icecream itself using the flags internally, and that triggers a GCC bug. Your patch seems to be supposed to handle a case when the user explicitly specifies it, but the solution cannot be to just silently drop the flag, as that would change what the compiler does.

deriamis:
> What is the actual use case you're trying to solve? This doesn't help with #451, as that one is about icecream itself using the flags internally, and that triggers a GCC bug. Your patch seems to be supposed to handle a case when the user explicitly specifies it, but the solution cannot be to just silently drop the flag, as that would change what the compiler does.

I think there must have been some confusion as to what was happening in the original issue. I've tested this behavior many times, and using `ICECC_REMOTE_CPP=1` doesn't cause any problems until I add `-fdirectives-only` to my local command line. This change actually fixes the original problem by preventing the flags from being used where they should not have been used in the first place.

The problem is that source preprocessed with `-fdirectives-only` is passed to the compiler and the compiler *also* has the flag passed in from the local command line. The `-fdirectives-only` flag is supposed to only be used once. If you do separate preprocessing and compiling steps, it should only be passed to the preprocessor.

The reason the change drops the flag entirely is because I was addressing the remote preprocessing case, which implies separate preprocessing and compilation. Nether `-fdirectives-only` nor `-frewrite-includes` makes much sense to use unless you are doing that anyway. I could limit dropping the flags to remote compilation only, or I could make the solution a bit more general by putting `-fpreprocessed` in place of `-fdirectives-only` in the compiler args, either which should preserve expected compiler behavior. Would either of those work for you?

llunak:
> I think there must have been some confusion as to what was happening in the original issue. I've tested this behavior many times, and using `ICECC_REMOTE_CPP=1` doesn't cause any problems until I add `-fdirectives-only` to my local command line. This change actually fixes the original problem by preventing the flags from being used where they should not have been used in the first place.

The confusion is most likely on your part. I've just tested with GCC 9.3.1 and the problem still exists, both with and without using icecc. So again, what *exactly* is the problem you're trying solve?

deriamis:
> > I think there must have been some confusion as to what was happening in the original issue. I've tested this behavior many times, and using `ICECC_REMOTE_CPP=1` doesn't cause any problems until I add `-fdirectives-only` to my local command line. This change actually fixes the original problem by preventing the flags from being used where they should not have been used in the first place.
> 
> The confusion is most likely on your part. I've just tested with GCC 9.3.1 and the problem still exists, both with and without using icecc. So again, what _exactly_ is the problem you're trying solve?

Sorry for the delay in response. I am indeed confused, and much of the reason has to do with the fact that none of this is well-documented and the GCC devs have not investigated the behavior. Also, the original bug report has nothing to do with `-fdirectives-only` behavior as much as it does differences in preprocessing behavior between using `cpp` and `gcc -E`.

I really can't go into the exact problem I am seeing (trust me, it's very complicated), but I spent some time trying to get a good repro for this issue to make what's happening obvious. It's changed what I think is happening here.

First, this is the test code (similar to the original GCC bug, but without the extra issues):
```c++
#define DECLARE const auto
#include <string>
#include <iostream>
int main()
{

    DECLARE vert = R"(#hello)";
    std::cout << vert << std::endl;
    return 0;
}

```
This always succeeds:
```bash
g++ -fdirectives-only test.cpp -o test
```

The actual problem is caused by passing source preprocessed with `-fdirectives-only` to the compiler where `-fdirectives-only` has not been specified:
```bash
g++ -fdirectives-only -E test.cpp -o test.ii #prepares the preprocessed source
g++ -c test.ii -o test #fails
g++ -fpreprocessed -c test.ii -o test #fails
```

This is the correct way to build with `-fdirectives-only`:
```bash
g++ -fdirectives-only -E test.cpp -o test.ii #prepares the preprocessed source
g++ -fdirectives-only -c test.ii -o test
g++ -fpreprocessed -fdirectives-only -c test.ii -o test #also works, but not really necessary
```

What I think is probably happening is that icecream is compiling source preprocessed with `-fdirectives-only` without passing the same flag to the compiler. Fixing the original problem is going to require fixing both that *and* adding a choice to use `cpp` over `gcc -E` when preprocessing. I'm going to dive into the source this weekend and see what I can come up with.

ossilator:
-fpreprocessed is implied when compiling an .ii file, which is why you don't see a difference.

let's have a look at the cases:
- when called as gcc -E, icecc does a local pass-through, so the case is irrelevant
- so is any other case that causes local pass-through - it's simply none of icecc's business
- when distributing compilation of raw sources, icecc splits off pre-processing and by default adds -fdirectives-only to both passes. the user may have passed -fdirectives-only, but then they are just being redundant. and if the user set ICECC_REMOTE_CPP=0 but passed the flag (which is passed on to both passes), they are just reproducing icecc behavior.
- when distributing compilation of pre-processed sources, icecc skips pre-processing, and needs to take -fdirectives-only passed by the user at face value, as it can't know otherwise

passing -fdirectives-only to a _compilation_ of raw sources is semantically meaningless, so no conclusions either way can be drawn from the results.

the conclusion is that pruning -fdirectives-only is pointless or even just plain wrong in every single case.

you might be able to work around the original issue in some way, but the patch and its underlying logic are a dead end.

deriamis:
> -fpreprocessed is implied when compiling an .ii file, which is why you don't see a difference.
> 
> let's have a look at the cases:
> 
>     * when called as gcc -E, icecc does a local pass-through, so the case is irrelevant
> 
>     * so is any other case that causes local pass-through - it's simply none of icecc's business
> 
>     * when distributing compilation of raw sources, icecc splits off pre-processing and by default adds -fdirectives-only to both passes. the user may have passed -fdirectives-only, but then they are just being redundant. and if the user set ICECC_REMOTE_CPP=0 but passed the flag (which is passed on to both passes), they are just reproducing icecc behavior.
> 
>     * when distributing compilation of pre-processed sources, icecc skips pre-processing, and needs to take -fdirectives-only passed by the user at face value, as it can't know otherwise
> 
> 
> passing -fdirectives-only to a _compilation_ of raw sources is semantically meaningless, so no conclusions either way can be drawn from the results.
> 
> the conclusion is that pruning -fdirectives-only is pointless or even just plain wrong in every single case.
> 
> you might be able to work around the original issue in some way, but the patch and its underlying logic are a dead end.

Please let me be clear - I am not proposing that my original patch be taken. I have been doing a lot more investigating since then, and I am not 100% clear on what exactly in Icecream might be contributing to the problem I am seeing, if anything at all. What's happening in my build is that Icecream is separately being called as a preprocessor and then a compiler, and the places where `-fdirectives-only` is and is not being used by both tools is causing a lot of confusion. It looks almost exactly like the original problem, which is why I brought it up.

As I said, I am still investigating, and I'm currently working with the first tool in the toolchain. I'll revisit this if there is still a problem.

deriamis:
I have some interesting results. I was actually doing this investigation with the initial impression that there were separate and interacting bugs in ccache and Icecream, and I've also been maintaining a thread on that project about our observations. Rather than cross-post it here, I'll refer you there:

https://github.com/ccache/ccache/issues/617

From the perspective of the Icecream project, what is happening is that the GCC bug was apparently noticed but not recognized as such. From the comments in the source, I gather that GCC's `-fdirectives-only` is known to be troublesome in ways that clang's `-frewrite-includes` is not, and efforts were made to retry preprocessing without the flag if it fails.

To make things clearer, the actual bug in GCC is that preprocessing behavior differs by how the preprocessor is invoked. The preprocessor binary (`cpp`, which is what gets invoked by `gcc -E`) will fail to preprocess the source in #451 because it contains an invalid preprocessor directive. However, the internal preprocessor will just silently remove the invalid preprocessor directive and blithely continue on.

It looks like Icecream tries to work around the bug (the behavior noticed in #451) by dropping `-fdirectives-only` from both the preprocessor and compiler arguments whenever preprocessing fails. Unfortunately, the flag also gets dropped for *any* reason preprocessing fails, including failing to contact the remote host. The behavior we have been observing, and what I have been incorrectly attempting to address without realizing it, is that Icecream is dropping the `-fdirectives-only` flag when it should not have been. The workaround as implemented seems to be incorrect.

I'm going to work on a new patch that should correct the workaround and make what I am describing much more clear. I'll try to get it done by the end of the week if life doesn't interfere.

deriamis:
Based on the discussion we're having here, I'm closing this PR out and opening a new one that's targeted to what we have figured out since. I'll be putting an initial PR together for it today as well as commenting on the GCC bug with our findings.

PR #560, opened 2020-09-14
Change fd2channel to fd2clien in daemon.

kalin1993:
This change reduce number of lookups in clients map and allows to make Client class the only owner of MsgChannel(for particular client)

### DIFF
--- a/daemon/main.cpp
+++ b/daemon/main.cpp
@@ -260,16 +260,6 @@ public:
         return nullptr;
     }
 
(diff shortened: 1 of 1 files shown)


deriamis (approved):
Sorry for the long delay in review. The tests on this PR were (a lot) stale, so I checked it out and ran the tests locally. I'm convinced that this is good to go as-is. Thanks for the PR!

PR #561, opened 2020-09-29
fix daemon setting wrong memory limitation

cinakyn:
On macOS + clang environment, Clustering much more than 60 jobs makes kill my macOS (It even terminate my machine.)
I found that the daemon had been logging weird message about memory limitation.

```
[70527] 2020-09-29 09:28:11: Compile job memory limit set to 3459279 megabytes
[70496] 2020-09-29 18:28:11: handle connection returned 70529
[70496] 2020-09-29 18:28:11: request for job 52
```
That means the daemon set limitation to almost 34 terabytes.

I have found that the function `unsigned int culateMemLoad(...)` make free memory not in kilobytes but bytes.
In addition to that, the daemon limits max child counts into at most 4. But my machine has 20 children.

So this PR is trying to fix above two problems.

### DIFF
--- a/daemon/load.cpp
+++ b/daemon/load.cpp
@@ -240,7 +240,7 @@ static unsigned long int scan_one(const char *buff, const char *key)
 }
 #endif
 
-static unsigned int calculateMemLoad(unsigned long int &NetMemFree)
+static unsigned int calculateMemLoad(unsigned long long &NetMemFree)
 {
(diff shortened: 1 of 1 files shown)


cinakyn:
I think finding why the one test has failed is out of my ability in this time...

deriamis (approved):
I don't think the test failure was your fault. This LGTM. Thanks for the PR!

PR #562, opened 2020-10-08
fix "call_cpp" could be executed more time than max_cpu

cinakyn:
On macOS + clang environment, Clustering much more than 60+ jobs makes kill my macOS (It even terminate my machine.)
I have found "call_cpp" is executed much more time than machine's cpu core count.
I guess it might came from #527

### Environment
- 5 x (12 core mac mini) cluster
- macOS catalina

### Issue
My machine is forcefully terminated when 30 ~ 60 compile job just had distributed to cluster.

### Reason (I guess)
1. There're 60 clients totally. 12 icecc client have sent `CompileFileMsg` and got `dcc_lock_host` there's no remain lock now.
2. After very short time, they execute `call_cpp` and release the lock
https://github.com/icecc/icecream/blob/19680ed605f13a5c7b598587fcfcdfd9ac8022ec/client/remote.cpp#L290-L298
3. Another client can take the lock and execute 13th `call_cpp`
4. Eventually all 60 clients run `call_cpp` simultaneously as repeating step2 ~ step 3 very quickly.
5. Boom

I have moved some lock to other positions and checked that this PR solving the issue.
Now I'm using 80 clients for compiling my projects.
Thank you.

### DIFF
--- a/client/remote.cpp
+++ b/client/remote.cpp
@@ -270,7 +270,7 @@ static void check_for_failure(Msg *msg, MsgChannel *cserver)
 
 // 'unlock_sending' = dcc_lock_host() is held when this is called, temporarily yield the lock
 // while doing network transfers
-static void write_fd_to_server(int fd, MsgChannel *cserver, bool unlock_sending = false)
+static void write_fd_to_server(int fd, MsgChannel *cserver)
 {
     unsigned char buffer[100000]; // some random but huge number
     off_t offset = 0;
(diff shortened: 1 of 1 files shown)


deriamis (approved):
At MongoDB, we have seen instances of Icecream killing local hosts, and now I think you have pointed out the reason why. This LGTM as far as the changes go, and I can confirm with some local testing that it at least doesn't lead to an obvious regression. I'm going to include the patch in MongoDB's private Icecream package to give it some bake time just to be sure, but I'm convinced there's no reason not to merge it now.

deriamis:
@llunak @HenryMiller1 We had some more instances of our developer workstations getting wedged by Icecream tasks. I have a patched package out for testing now to see if this commit fixes it. If it does, how would you feel about releasing 1.4.2?

PR #579, opened 2021-03-29
Fix autogenerated environments if /bin/cc is clang

nabijaczleweli:
аnd a depessimisation

See commit message for reasoning

### DIFF
--- a/compilerwrapper/compilerwrapper.cpp
+++ b/compilerwrapper/compilerwrapper.cpp
@@ -62,24 +62,33 @@ int main(int argc, char *argv[])
     fprintf(stderr, "\n");
 #endif
     bool isclang = argc >= 2 && strcmp(argv[1], "clang") == 0;   // the extra argument from icecream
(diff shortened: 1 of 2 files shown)


nabijaczleweli:
I mean, the alternative would be a goto instead, I guess? But there weren't any gotos in C++ code, so I didn't want to add any. I wrapped it in parens for the express reason of silencing this warning

coolo:
Your goal is to call the block twice, no? Make it a function?

nabijaczleweli:
You're right, of course. Updated

nabijaczleweli:
Bumping after a month. Is this missing something?

HenryMiller1 (approved):

HenryMiller1:
The CI failures are concerning, but I don't understand what they mean.  Can you rebase this to the latest to see if it  works?

nabijaczleweli:
rebased

nabijaczleweli:
aaand it passes cleanly

PR #595, opened 2022-02-25
Refactor pick_server and add new scheduler algorithms

deriamis:
The classic Icecream scheduler doesn't work well in all environments, especially when deployed into the cloud. To remedy that, `pick_server` is refactored, and several new scheduling algorithms are added as options:
  * random
  * round_robin
  * least_busy
  * fastest

The `fastest` algorithm is the "classic" algorithm.

I've also fixed up a few parts of the `fastest` algorithm and server selection in general to be a bit more obvious in their function. In particular, the original weight assumed that there were at least 1000 nodes in the cluster. This has been changed to scale along with the size of the cluster.

### DIFF
--- a/scheduler/scheduler.cpp
+++ b/scheduler/scheduler.cpp
@@ -47,6 +47,7 @@
 #include <cassert>
 #include <fstream>
 #include <string>
+#include <limits>
 #include <stdio.h>
 #include <pwd.h>
 #include "../services/comm.h"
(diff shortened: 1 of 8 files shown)


scheduler/scheduler.cpp
```diff
@@ -624,72 +626,104 @@ static string envs_match(CompileServer *cs, const Job *job)
...
+#endif
+                return false;
+            }
 
-    /* consistency checking for now */
-    for (list<CompileServer *>::iterator it = css.begin(); it != css.end(); ++it) {
-        CompileServer *cs = *it;
+            /* Don't use non-chroot-able daemons for remote jobs.  XXX */
+            if (!cs->chrootPossible() && cs != job->submitter()) {
+                trace() << cs->nodeName() << " can't use chroot\n";
```
dmoody256, line 658: should this be under debug?

scheduler/scheduler.cpp
```diff
@@ -624,72 +626,104 @@ static string envs_match(CompileServer *cs, const Job *job)
...
+            }
 
-        list<Job *> jobList = cs->jobList();
-        for (list<Job *>::const_iterator it2 = jobList.begin(); it2 != jobList.end(); ++it2) {
-            assert(jobs.find((*it2)->id()) != jobs.end());
-        }
-    }
+            // Check if remote & if remote allowed
+            if (!cs->check_remote(job)) {
+                trace() << cs->nodeName() << " fails remote job check\n";
```
dmoody256, line 664: debug define?

scheduler/scheduler.cpp
```diff
@@ -624,72 +626,104 @@ static string envs_match(CompileServer *cs, const Job *job)
     return string();
 }
 
-static CompileServer *pick_server(Job *job)
+static list<CompileServer *> filter_ineligible_servers(Job *job)
 {
+    list<CompileServer *> eligible;
+    std::copy_if(
+        eligible.begin(),
```
dmoody256, line 633: didn't this list just get created and so therefore its empty? Shouldn't you be iterating some populated list and back inserting into this new empty list?

scheduler/scheduler.cpp
```diff
@@ -624,72 +626,104 @@ static string envs_match(CompileServer *cs, const Job *job)
     return string();
 }
 
-static CompileServer *pick_server(Job *job)
+static list<CompileServer *> filter_ineligible_servers(Job *job)
 {
+    list<CompileServer *> eligible;
```
dmoody256, line 631: I don't see a reason why list should be used over vector here. Any arguments against a vector?

scheduler/scheduler.cpp
```diff
@@ -624,72 +626,104 @@ static string envs_match(CompileServer *cs, const Job *job)
...
+    return eligible;
+}
 
-    for (map<unsigned int, Job *>::const_iterator it = jobs.begin();
-            it != jobs.end(); ++it) {
-        Job *j = it->second;
+static CompileServer *pick_server_random(list<CompileServer *> &eligible)
+{
+    auto iter = eligible.cbegin();
+    std::advance(iter, random() % eligible.size());
```
dmoody256, line 676: if this was a vector it would be a bit faster to pick a random than iterating

scheduler/scheduler.cpp
dmoody256: should you be iterating css? That seems like some kind of global list just reading the diffs, not sure where its coming from though.

scheduler/scheduler.cpp
```diff
@@ -624,72 +626,104 @@ static string envs_match(CompileServer *cs, const Job *job)
...
+    return *iter;
+}
 
-        if (j->state() == Job::COMPILING) {
-            CompileServer *cs = j->server();
-            list<Job *> jobList = cs->jobList();
-            assert(find(jobList.begin(), jobList.end(), j) != jobList.end());
+static CompileServer *pick_server_least_busy(list<CompileServer *> &eligible) {
+    eligible.sort([] (const CompileServer *a, const CompileServer *b) {
```
dmoody256, line 681: whats the reason for sorting? It doesn't seem the list gets reused.

scheduler/scheduler.cpp
```diff
@@ -742,26 +741,18 @@ static CompileServer *pick_server(Job *job)
...
+            weight_factor = std::numeric_limits<uint8_t>::max() / weight_limit;
         }
 
-        /* Distribute 5% of our jobs to servers which haven't been picked in a
-           long time. This gives us a chance to adjust the server speed rating,
-           which may change due to external influences out of our control. */
-        if (!cs->lastPickedId() ||
-            ((job->id() - cs->lastPickedId()) > (20 * css.size()))) {
+        if (weight_factor > 0 && (!cs->lastPickedId() ||
+            ((job->id() - cs->lastPickedId()) > (weight_factor * eligible.size())))) {
```
dmoody256, line 755: it's not easily identifiable what exactly this condition is doing, could you add a comment explaining it. For example, what do job ids have to do with the number of eligible servers?

scheduler/scheduler.cpp
deriamis: Ordinarily, yes, I would prefer a `vector`. However, I was keeping the data type the same during the refactor. I'm also trying to change as little code as possible to get the job done and following the surrounding conventions. Keep in mind that this is C++11 - we're not dealing with "modern" C++ standards here.

scheduler/scheduler.cpp
deriamis: Yeah, I changed how this worked when I realized I didn't have what I wanted because this is C++11. I was trying to get it to compile and forgot to go back and change what I was iterating. It should be `css`, which is a static declared at the top of the file.

scheduler/scheduler.cpp
deriamis: Nope. `trace` is just the debug log stream. The only time they put anything behind a debug define is when they're doing something expensive.

scheduler/scheduler.cpp
deriamis: Yeah, I know it would. The original data structure is a list of pointers, though, so that's what I'm doing here.

scheduler/scheduler.cpp
deriamis: I'm returning the first element of the sorted list.

scheduler/scheduler.cpp
deriamis: Job id's are a monotonically increasing sequence. By looking for the last job id a server did, I can tell when it was last used. This is how another (unchanged) part of the code does something similar.

scheduler/scheduler.cpp
```diff
@@ -624,72 +626,104 @@ static string envs_match(CompileServer *cs, const Job *job)
...
+                selected = cs;
             }
         }
-
-        return nullptr;
     }
+    return selected;
+}
 
+static CompileServer *pick_server_fastest(Job *job, list<CompileServer *> &eligible, uint8_t weight)
```
RedBeard0531, line 720: While you are here, you should deprioritize any server that is more than half full. Maybe by reducing its assumed speed by 50 or even 75 percent. This is based on the assumption that essentially all modern systems use SMT/HyperThreads, so only half of your "cores" are real, and adding more threads past that, while still a marginal improvement, doesn't help quite as much as adding a real core.

A while back, I deployed this tweak[1] to a modified scheduler in the MongoDB NYC office and IIRC (it was many years ago...) got something like a 25% speedup on clean build times. Do we have any ability to test modified scheduler algorithms on our new cluster?

[1] I think I actually used a very strong discounting of `speed *= 0.1` to really try hard to fill up physical cores before touching logical ones, and also to get the secondary benefit of not consuming all cores on anyone's desktops unless the entire cluster was more than half full (which rarely happened at that point because we had >1000 cores, and most compiles used 2-300).

deriamis: The algorithms I'm adding are "better" in the sense because they assume all hosts are essentially equivalent but may be receiving different jobs that affect its perceived speed. This is the case for us and anyone else using hosts in the cloud, but it might not be true of in-house situations. I don't like this algorithm at all, but I'm trying to keep the defaults as similar to what they were before for now.

In all honesty, this is perhaps a large change for a point release, but I'm trying to keep it as small as I can (for now) while adding what we need. What you're suggesting is closer to a major version change - which I would like to see happen sooner rather than later, but I want something incremental first.

scheduler/scheduler.cpp
HenryMiller1: Which is as it should be - if there are problems it is a lot better if we can say enable logs and send them to us vs rebuild in debug first.

scheduler/scheduler.cpp
HenryMiller1: recafctor to vvector is a good idea, but shouldn't be done in this pull request.

scheduler/scheduler.cpp
```diff
@@ -908,7 +1054,7 @@ static time_t prune_servers()
     return min_time;
 }
 
-static bool empty_queue()
+static bool empty_queue(SchedulerAlgorithmName scheduler)
```
HenryMiller1, line 1057: calling this variable scheduler doesn't make sense.  Can it be renamed to algorithm?  Or even schedulerAlgroithm

HenryMiller1 (changes requested):
I don't see documentation, this new option will be very important for users to play with so it needs to be clearly documented. 

Make sure that you note that the "Fastest" may not be the fastest in the real world and users should try all options to see what works best.  I'm interested in making a different algorithm the default in the  future after we get more experience with them in the real world.

scheduler/scheduler.cpp
HenryMiller1: It does seem like future optimizations of this algorithms could account for the number of CPUs that each server has, but that need not be part of this.  (and if made might be a new user selected able algorithm leaving these in place)

deriamis: @HenryMiller1 That's essentially what the "least busy" algorithm is trying to do, except with the maximum number of configured jobs substituting for the number of CPUs available. Here, I think "fastest" can and should mean what it always has - the host we expect will complete the task soonest, which isn't necessarily the least-busy one.

scheduler/scheduler.cpp
deriamis: Yes, I think this is one I missed in my search-and-replace. I'll fix it up in the next commit.

deriamis:
> I don't see documentation, this new option will be very important for users to play with so it needs to be clearly documented.
> 
> Make sure that you note that the "Fastest" may not be the fastest in the real world and users should try all options to see what works best. I'm interested in making a different algorithm the default in the future after we get more experience with them in the real world.

Documentation was my next step after getting some feedback on how it would operate. Given your positive initial response, I'll work on the documentation next.

deriamis:
@HenryMiller1 I've added the documentation you requested. PTAL. Please note, this PR also has changes from #596 and #597 in it, which were needed for testing, so if you would not mind taking a look at those as well, I would appreciate it. Thanks!

HenryMiller1:
Does anyone understand the circle_ci build error? Looks like it is happening everywhere, so probably not this, but I still don't like accepting code that doesn't pass Ci.

HenryMiller1 (approved):

deriamis:
> Does anyone understand the circle_ci build error? Looks like it is happening everywhere, so probably not this, but I still don't like accepting code that doesn't pass Ci.

I took a quick look, and from what I can tell it's happening because the version of `gm4` installed from `pkg` is no longer compatible with FreeBSD 12 (which is EOL). I don't know anything beyond that, nor do I have the access to fix it.

scheduler/scheduler.h
```diff
@@ -27,4 +27,68 @@
 // Values 0 to 3.
 #define DEBUG_SCHEDULER 0
 
+// The weight the "fastest" scheduler places on using a
+// server with recent statistics over one that has not
+// been used for a while. Values 0 - 128, with higher
+// values meaning the "fastest" servers are used more
+// often.
+#define STATS_UPDATE_WEIGHT 120
```
dmoody256, line 35: this used to be adjustable from command line in previous commit right? Just curious why it changed?

deriamis: I was starting to get too far into the weeds on how to parse the command-line options when what I wanted to do was offer alternatives to the original algorithm. It's possible such an option will return when it has relevance to more than one algorithm, or when we have an idea of how changing the value might affect the results.

dmoody256 (approved):

PR #596, opened 2022-02-26
Fix tests

deriamis:
I tried running the tests, and I was getting consistent failures due to changes in the compilers and in `icecc`/`iceccd` that haven't been reflected in the tests yet. This commit just fixes up the tests so they run properly and pass with recent compilers and source.

PR #597, opened 2022-02-26
Add llvm-symbolizer to environment tarball

deriamis:
We weren't adding `llvm-symbolizer` to the environment tarball even though we assume it is present in the tests. It really should be put in there when available, or sanitizer output coming from a remote job will be hard to decipher.


HenryMiller1:
This can be closed, right?

deriamis:
Yep, merged in commit [c87373f](https://github.com/icecc/icecream/commit/c87373f033fd81aca959776652f074b87f170152) as part of #596.

PR #601, opened 2022-04-30
Refactor the `Msg` class

deriamis:
This is a refactor of the `Msg` class that gets rid of the need for `MsgType`, which was a separate, non-scoped `enum`. Having the non-scoped `enum` around meant we had bare identifiers everywhere. I'm using an "enhanced `enum`" here (the kind we used for  `SchedulerAlgorithmName`) to do the refactor.

The benefits of the refactor are that we no longer need a separate variable to hold the `enum` value anymore because we can just use the object itself, and as a result it's a lot easier to output the actual name of the message in logs. In addition, having a scoped `enum` means the object is more self-container, and it will be easier to do other refactors.

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -10,12 +10,12 @@ task:
     image_project: freebsd-org-cloud-dev
     platform: freebsd
   matrix:
-    - name: 13.0-RELEASE
+    - name: 13.1-RELEASE
       compute_engine_instance:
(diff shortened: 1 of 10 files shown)


deriamis:
Sorry for the mess with all the force commits. I was working on something else entirely and realized too late I needed to split this out. The next PR will be a lot better.

PR #610, opened 2022-12-13
Fix the monitor GetCS and JobDone messages not having proper type

Zalewa:
Broken by the refactor at c0f9d537bd47e3326e1126444267cf43bd59696f


krf:
Can confirm this is also fixing Icemon with the new Msg API.

PR #620, opened 2023-06-21
avoid integer overflow for cache-limit

michaelolbrich:
Without this, "mb * 1024 * 1024" overflows for mb >= 2048. The result is then cast into size_t, which is unsigned and usually a 64-bit integer. The end result is, that the cache is now basically unlimited.

Fix this by making mb a size_t as well to avoid the overflow.

### DIFF
--- a/daemon/main.cpp
+++ b/daemon/main.cpp
@@ -2349,7 +2349,7 @@ int main(int argc, char **argv)
                     int mb = atoi(optarg);
 
                     if (!errno) {
-                        cache_size_limit = mb * 1024 * 1024;
+                        cache_size_limit = mb * 1024L * 1024;
                     }
                 } else {
                     usage("Error: --cache-limit requires argument");


daemon/main.cpp
```diff
@@ -2346,7 +2346,7 @@ int main(int argc, char **argv)
             } else if (optname == "cache-limit") {
                 if (optarg && *optarg) {
                     errno = 0;
-                    int mb = atoi(optarg);
+                    size_t mb = atoi(optarg);
```
ossilator, line 2349: i don't think this is the right thing to do. it doesn't match the return value of atoi(), and it obfuscates the actual problem.

instead, you should add L to the 1024's.

michaelolbrich: Done.

ossilator (commented):
"operate_s_" in the commit message.
also, the statement isn't true on 32-bit systems.

hmm. so pedantically, this is all wrong anyway. it shouldn't be size_t, but off64_t or some such. but i suspect that the code is full of such bugs anyway, so whatever. "nobody" still builds on 32-bit machines anyway.

some error handling would be nice, though. but that's a separate thing.

PR #626, opened 2023-11-12
Revert "Refactor Msg class"

stefanu21:
This reverts commit c0f9d537bd47e3326e1126444267cf43bd59696f.

This is a break in the api. Projects like icemon no longer works.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -50,8 +50,6 @@ tests/testargs.log
 tests/testargs.trs
 tests/test-setup.sh
 tests/listing.txt
-unittests/testargs
-unittests/testargs.trs
 # others
 compile_commands.json
 .vs*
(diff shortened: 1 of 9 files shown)


ossilator:
this feels very wrong to me.
instead, the internal api should be properly versioned, and icemon (the only external consumer?) should have the required #ifdefs.

stefanu21:
yes you are right but the mistake has already happened that the api is not versioned correctly. nevertheless there is no compelling reason to change the api like this and make the clients unusable. after all, this is just cosmetics. if it would be about projects that would be developed much further ok but neither icecc nor icemon has done anything substantial in the last years except the new scheduler modis...

icew4y (approved):

krf:
I think this is no longer needed. We've made the necessary changes in Icemon.

PR #627, opened 2023-12-06
Fix icecc-create-env in OpenSUSE Tumbleweed

jimis:
There the binaries depend on libraries of different minor version than the installed library, which is found at runtime via symlinks.

For example, ldd /usr/bin/clang-17 reports among other lines:

  libz.so.1 => /lib64/glibc-hwcaps/x86-64-v3/libz.so.1.3 (0x00007f3fdc8ff000)

Before the patch, icecc-create-env /usr/bin/clang-17 would report:

  adding file /lib64/libz.so.1.3=/usr/lib64/libz.so.1.3

The executable wouldn't run because it can't find libz.so.1. After the patch:

  adding file /lib64/libz.so.1=/usr/lib64/libz.so.1.3

And the executable runs fine.

### DIFF
--- a/client/icecc-create-env.in
+++ b/client/icecc-create-env.in
@@ -105,6 +105,11 @@ add_file ()
   is_add_file_duplicate "$name" && return
   add_file_duplicates="$add_file_duplicates $name"
 
+  # If the target name is not a full path, then it's a symlink in the same dir as "path".
+  # So prepend the (unresolved) path's directory to the name.
+  if [ $(dirname "$name") = "." ]; then
+      name="$(dirname "$path")/$name"
+  fi
   path=$(resolve_path $path)
   name=$(convert_path_cdup $name)
   if test -n "$stripprefix"; then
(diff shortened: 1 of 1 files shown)


kdave:
Thank you so much, so glad I found this. Figuring out why the worker hosts don't compile anything is painful, on openSUSE it's indeed due to the libz.1 and symbol mismatches. Your fixed script works for me, hopefully it can be merged.

kdave:
There's probably missing case to handle zlib-ng-compat package, the file in the environment tar.gz is `usr/lib64/zlib-ng-compat/libz.so.1` and also fails to resolve on the worker host.

kdave:
So, removing `zlib-ng` and its compat packages works, `libz1-x86-64-v3` is handled by the fixed script.

HenryMiller1:
#653 seems to fix the same in a simpler way.  Let me know if there are more issues

Issue #382, opened 2018-02-28
Error 14 for large compiler output (Around 1 Megabyte)


MaxSagebaum:
We recognized that build jobs fail on remote machines. The issue we identified is just the size of the generated output from the compiler. Around a file size of 1 Megabyte the remote job fails with the error 14. We created a small test program, that produces large amounts of warnings. The test program could be used to recreate the issue.

The warning is:
> test_icecc.cpp: In instantiation of 'void func(int) [with int dim = 0]':
> test_icecc.cpp:6011:1:   required from here
> test_icecc.cpp:4:33: warning: unused parameter 'arg' [-Wunused-parameter]

This warning is repeated 6000 times until the file size is reached.

The full log for the ungraceful exit is:
> ICECC[28001] 16:06:34: connected to /var/run/icecc/iceccd.socket
> ICECC[28001] 16:06:34: env: x86_64 '/home/msagebaum/Applications/42b4b3fed6af9434d65174165c0f4e83.tar.gz'
> ICECC[28001] 16:06:34: test_icecc.cpp compiled 1 times on x86_64
> ICECC[28001] 16:06:34: Have to use host 131.246.20.79:10245 - Job ID: 6869 - env: x86_64 - has env: true - match j: 0
> ICECC[28001] 16:06:34: connected to 131.246.20.79
> ICECC[28001] 16:06:34: <send compile_file>
> ICECC[28001] 16:06:34: </send compile_file: 0ms>
> ICECC[28001] 16:06:34: <write_server_cpp from cpp>
> ICECC[28001] 16:06:34: sent 24714 bytes (29%)
> ICECC[28001] 16:06:34: </write_server_cpp from cpp: 32ms>
> ICECC[28001] 16:06:34: <wait for cpp>
> ICECC[28001] 16:06:34: </wait for cpp: 0ms>
> ICECC[28001] 16:06:34: <wait for cs>
> ICECC[28001] 16:06:36: !read_a_bit 2
> ICECC[28001] 16:06:36: !wait_for_msg()
> ICECC[28001] 16:06:36: </wait for cs: 1198ms>
> ICECC[28001] 16:06:36: timeout <= 0
> ICECC[28001] 16:06:36: no message type
> ICECC[28001] 16:06:36: got exception Error 14 - error reading message from remote (131.246.20.79)

For the working version the log is:
> ICECC[28079] 16:09:04: connected to /var/run/icecc/iceccd.socket
[...]

PR #650, opened 2026-01-18
fix: don't fail on large messages (for projects like LLVM)

git-bruh:


services/comm.cpp
```diff
@@ -263,9 +263,7 @@ bool MsgChannel::update_state()
             (*this) >> inmsglen;
 
             if (inmsglen > MAX_MSG_SIZE) {
-                log_error() << "received a too large message (size " << inmsglen << "), ignoring" << endl;
-                set_error();
-                return false;
+                log_warning() << "received a too large message (size " << inmsglen << ")" << endl;
```
HenryMiller1, line 266: Changing to a warning and continuing is the wrong answer.  Maybe we should change MAX_MSG_SIZE, or better you make it configurable. 

This change will confuse future maintainers who are going to think MAX_MSG_SIZE (perhaps + 1) is the largest buffer they need.  Even if this works today someone in the future may make a change and break this since there are no tests we won't even find out that this obscure case is broken.

I know icecc says only use on a local trusted network, but that isn't an excuse to allow any attacker to send any amount of data.  For security reasons alone this is rejected.  We might not be secure code, but that isn't an excuse to introduce more potential security issues.

HenryMiller1 (changes requested):
The max size can be configurable, but we need to reject anything larger than max size.

HenryMiller1:
Your base branch is closed. Changes to master only are allowed.  Even if this is something for already released code, we still need this in master as well if we are to do it.

services/comm.cpp
git-bruh: yeah it makes more sense to have it be configurable

however one thing comes to mind - the max message size is handled independently on each machine, it might be a bit cumbersome to ensure that they all use the same value, so bumping up the MAX_MSG_SIZE to a value that can reasonably be hit in large projects like llvm, chromium, etc. might be better from a usability standpoint. I'm personally fine with either since this is a fairly niche case (the only other mention I found is https://github.com/icecc/icecream/issues/382)

PR #651, opened 2026-01-19
support niceness for local bulids

jp-embedded:
icecc priorotizes jobs based on the process's niceness. But not for local jobs. This PR fixes that.

### DIFF
--- a/client/main.cpp
+++ b/client/main.cpp
@@ -121,6 +121,15 @@ static void icerun_show_usage()
         "\n");
 }
 
+static int get_niceness()
+{
+    errno = 0;
+    int niceness = getpriority( PRIO_PROCESS, getpid());
+    if( niceness == -1 && errno != 0 )
+        niceness = 0;
+    return niceness;
+}
+
 volatile bool local = false;
 
 static void dcc_client_signalled(int whichsig)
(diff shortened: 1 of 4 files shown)


HenryMiller1:
Last I heard nice was almost meaningless on linux and other schedulers.  If it isn't we should use very low priority (high nice) to let interactive things work.  This seems like it is mostly useless and something we should rip out.

jp-embedded:
The niceness of the linux processes might not be useful. Put the prioritization of the queued icecc jobs is. So we have this project where a lot of linker jobs gets queued because they can run locally only. These then blocks other local jobs (libraries) that are queued after the linking jobs, until these linker jobs are done. The result is a reduced number of parallel jobs since a lot of following jobs are depending/waiting for the libraries.

Now, icecc can prioritize which jobs to execute first which solves this problem. This execute-order is determined by reading the niceness of the process. 

So I don't think we should rip out this execute-order prioritization, but it would probably be better/more logical to provide this priority as an argument to icecc/icerun instead of reading it from its niceness. But since this is the way it was already made, I added this missing piece the same way so it now works for both remote and local jobs.

Now, we run icecc as a prefix like "icecc gcc ...." instead of using gcc as a symbolic link to icecc. So it would be easy enough for us to add an priority argument to icecc, but its not so straight forward when using a symbolic link. I guess that is why it is made to use the process niceness instead of an argument in the first place.

PR #652, opened 2026-01-19
fix: disable remote preprocessing if -imacros is passed

git-bruh:


HenryMiller1:
Make your change against master before I will look at at.

